PR #3, opened 2024-01-02
Yuzuriha [WIP, do not merge]

Mach131:
Implementing stuff for Yuzuriha and S6 in general, mostly just to figure out how things are set up. I haven't looked into remote_player and there are some bugs with the AI playing her (because of that?), but she seems to be working properly when the player controls her in an AI match


scenes/game/local_game.gd
```diff
@@ -27,9 +27,12 @@ var all_cards : Array = []
 var game_over : bool = false
 var game_over_winning_player : Player = null
 var active_strike : Strike = null
+var active_character_action : bool = false
 var active_exceed : bool = false
 var active_overdrive : bool = false
+var active_revert : bool = false
```
daniel-k-taylor, line 33: I think instead of having the revert action be a completely separate thing, we should have it be another character_action.
We'd have to make the definition support listing multiple character actions, but I think overall this will be cleaner on a couple fronts:
1) in game.gd, no special case logic for s6_revert, just support looking for an array of character actions and making buttons for them
2) in localgame/remotegame, adding a parameter to submit/do_character_action for which character action you're doing
3) In the definition file, the 2nd character action is just move a card from hand to gauge, revert, and then take another action.

This may also make things simpler for adding the AI features for this.
Also, the code you've added for active_character_action will be the same as what you needed for active_revert, so some reuse there too.

daniel-k-taylor: As for adding the array, it should be pretty easy to have get_character_action/can_do_character_action take as input an int, and then the code in those functions can check if it is an array or a single character action and do the appropriate thing.
I mention this just so we don't have to go update every character definition to be an array, though that actually wouldn't be that bad either so maybe just do that for code simplicity?

scenes/game/local_game.gd
```diff
@@ -166,6 +172,12 @@ class Strike:
 		if performing_player == initiator:
 			return initiator_wild_strike
 		return defender_wild_strike
+		
+	func get_player_strike_from_gauge(performing_player : Player) -> bool:
+		if performing_player == defender:
+			return true
```
daniel-k-taylor, line 178: Shouldn't this be false? The defender can't strike from gauge?

scenes/game/local_game.gd
```diff
@@ -166,6 +172,12 @@ class Strike:
...
+		# ensure that the strike from gauge wasn't invalidated
+		return initiator_set_from_gauge and !initiator_wild_strike
```
daniel-k-taylor, line 180: use "not" here instead of ! please

scenes/game/local_game.gd
```diff
@@ -1459,17 +1510,25 @@ func advance_to_next_turn():
...
 	return events
 
-func initialize_new_strike():
+func initialize_new_strike(performing_player : Player, opponent_sets_first : bool):
 	active_strike = Strike.new()
-	active_strike.strike_state = StrikeState.StrikeState_Initiator_SetEffects
 	active_strike.effects_resolved_in_timing = 0
+	
+	active_strike.opponent_sets_first = opponent_sets_first
+	if !opponent_sets_first:
```
daniel-k-taylor, line 1518: put the positive case here first "if opponent_sets_first"

scenes/game/local_game.gd
```diff
@@ -1483,33 +1542,16 @@ func continue_setup_strike(events):
...
-				# Queue any events so far, then empty this tally and call do_strike.
-				event_queue += events
-				events = []
-				do_strike(defender, reading_card.id, false, -1)
-				ask_for_response = false
-			else:
-				events += defender.reveal_hand()
-		if ask_for_response:
-			events += [create_event(Enums.EventType.EventType_Strike_DoResponseNow, defender.my_id, 0)]
+		if !active_strike.opponent_sets_first:
```
daniel-k-taylor, line 1546: positive case here first so you don't even need the !

scenes/game/local_game.gd
```diff
@@ -1521,7 +1563,55 @@ func continue_setup_strike(events):
 
 		# All effects resolved, move to next state.
 		active_strike.effects_resolved_in_timing = 0
-		events += begin_resolve_strike()
+		if !active_strike.opponent_sets_first:
```
daniel-k-taylor, line 1566: positive case first

scenes/game/local_game.gd
```diff
@@ -1521,7 +1563,55 @@ func continue_setup_strike(events):
...
+			# TODO: Potentially they can EX here.
+			# Queue any events so far, then empty this tally and call do_strike.
+			event_queue += events
+			events = []
+			do_strike(active_strike.defender, reading_card.id, false, -1, active_strike.opponent_sets_first)
+			ask_for_response = false
+		else:
+			events += active_strike.defender.reveal_hand()
+	if ask_for_response:
+		if !active_strike.opponent_sets_first:
```
daniel-k-taylor, line 1597: positive case first

scenes/game/local_game.gd
```diff
@@ -2171,7 +2279,7 @@ func handle_strike_effect(card_id :int, effect, performing_player : Player):
 		"powerup_damagetaken":
 			var power_per_damage = effect['amount']
 			var total_powerup = power_per_damage * active_strike.get_damage_taken(performing_player)
-			if total_powerup > 0:
+			if total_powerup != 0 and (total_powerup > 0) == (power_per_damage > 0):
```
daniel-k-taylor, line 2282: Can you rewrite this to not be a somewhat confusing if case?
It's okay to have more lines/if blocks, I just prefer if it is readable on sight as to what is going on.
I favor explicit behavior at a glance over brevity for stuff like this.

scenes/game/local_game.gd
```diff
@@ -2568,7 +2705,54 @@ func do_remaining_overdrive(performing_player : Player):
...
+	change_game_state(Enums.GameState.GameState_Boost_Processing)
+	while remaining_character_action_effects.size() > 0:
+		var effect = remaining_character_action_effects[0]
+		remaining_character_action_effects.erase(effect)
+		events += do_effect_if_condition_met(performing_player, -1, effect, null)
+		if game_state == Enums.GameState.GameState_PlayerDecision:
+			# Player has a decision to make, so stop mid-effect resolve.
+			break
+
+	if game_state != Enums.GameState.GameState_PlayerDecision:
```
daniel-k-taylor, line 2720: This probably will not work if we added something later with a character action that does something and then does a strike.
That puts the game in GameState_WaitForStrike.
Another case could be if a character action gives a bonus action, maybe something like Hilda's flip? That doesn't want to pass the turn.

I think for now just add in the strike condition here since that's such an obvious case, and we'll call it good. If we add more characters that need this we'll just update it then.

daniel-k-taylor: Oh, maybe the bonus actions part will already handled by the bonus actions code.

scenes/game/local_game.gd
```diff
@@ -2941,6 +3128,8 @@ func handle_strike_attack_cleanup(performing_player : Player, card):
 	var other_player = _get_player(get_other_player(performing_player.my_id))
 	if performing_player.is_set_aside_card(card.id):
 		events += [create_event(Enums.EventType.EventType_SetCardAside, performing_player.my_id, card.id)]
+	elif performing_player.strike_stat_boosts.discard_attack_on_cleanup:
```
daniel-k-taylor, line 3131: Maybe the seal case should go before this discard one, just in case something wants to seal and discard? It should probably get sealed.

scenes/game/local_game.gd
```diff
@@ -3359,7 +3549,8 @@ func do_boost(performing_player : Player, card_id : int, payment_card_ids : Arra
 	event_queue += events
 	return true
 
-func do_strike(performing_player : Player, card_id : int, wild_strike: bool, ex_card_id : int) -> bool:
+func do_strike(performing_player : Player, card_id : int, wild_strike: bool, ex_card_id : int,
+		opponent_sets_first : bool = false) -> bool:
```
daniel-k-taylor, line 3553: as mentioned, this parameter will need to be added to remotegame and the ai versions of this function.

scenes/game/local_game.gd
```diff
@@ -3369,14 +3560,22 @@ func do_strike(performing_player : Player, card_id : int, wild_strike: bool, ex_
 		if performing_player.my_id != get_other_player(active_turn_player):
 			printlog("ERROR: Strike response from wrong player.")
 			return false
+	elif game_state == Enums.GameState.GameState_Strike_Opponent_Set_First:
+		if performing_player.my_id != active_turn_player:
+			printlog("ERROR: Tried to strike but not current player")
+			return false
+		if !opponent_sets_first:
```
daniel-k-taylor, line 3567: use "not" here please

scenes/game/local_game.gd
```diff
@@ -3390,15 +3589,23 @@ func do_strike(performing_player : Player, card_id : int, wild_strike: bool, ex_
 	# Lay down the strike
 	match game_state:
 		Enums.GameState.GameState_PickAction, Enums.GameState.GameState_WaitForStrike:
-			initialize_new_strike()
-			active_strike.initiator = performing_player
+			if !opponent_sets_first:
```
daniel-k-taylor, line 3592: use not

scenes/menu/main_menu.gd
```diff
@@ -151,6 +151,8 @@ func update_char(label, portrait, char_id):
 	var display_name = "Random"
 	if char_id == "random_s7":
 		char_id = "random"
+	elif char_id == "random_s6":
+		char_id = "blazbluelogo2"
```
daniel-k-taylor, line 155: it looked like you added the s6 logo but this is still blazblue?

data/card_definitions.json
```diff
@@ -6852,383 +6852,1036 @@
...
 		"effects": [
 			{
 				"timing": "during_strike",
-				"effect_type": "do_not_remove_eddie"
+				"effect_type": "cannot_stun"
+			},
+			{
+				"timing": "hit",
+				"effect_type": "set_strike_x",
+				"source": "top_discard_power",
```
daniel-k-taylor, line 6875: is it obvious in the UI how this ends up getting chosen? I suspect not. I'd like to try to add something, or maybe just auto sort the spent gauge to pick the highest between them so the user doesn't even have to worry about it?

daniel-k-taylor (changes requested):
Really awesome changes so far!

data/decks/yuzu.json
```diff
@@ -0,0 +1,174 @@
...
+	"s6_revert_action": true,
+	"character_action_default": {
+		"gauge_cost": 0,
+		"force_cost": 0,
+		"effect": {
+			"effect_type": "gauge_from_hand",
+			"min_amount": 1,
+			"max_amount": 1,
+			"and": {
+				"condition": "min_cards_in_gauge",
```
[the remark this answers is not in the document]
daniel-k-taylor: I see you added some code to work around this.
I think that's fine for now, this was pretty janky behavior to begin with.
I'll have to think if I want to refactor that or just leave as is for now.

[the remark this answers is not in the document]
Mach131: Yee this is the part I first reached out about; this specific case does seem to work, but it's possible that there might be other weird combinations of effects that don't properly keep track of whether they're coming from character actions

scenes/game/game.gd
```diff
@@ -1700,6 +1729,15 @@ func _on_reveal_hand(event):
 		pass
 	return SmallNoticeDelay
 
+func _on_reveal_random_gauge(event):
+	var player = event['event_player']
+	var card_id = event['number']
+	spawn_damage_popup("Random Gauge Card!", player)
+	
+	#TODO: would be nice to display the card briefly
```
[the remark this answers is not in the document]
Mach131: Will look into this; do you have a good reference for temporarily moving it over? It's only being "selected" from gauge so I wanted to avoid actually moving it out of the gauge logically, but maybe if it was just on a timer somehow

scenes/game/local_game.gd
Mach131: Will work on this once I clean up some of the smaller stuff. What I think I may try to do is add support for multiple character actions, but keep in the "s6_revert_action" for the json file, then have a definition for the revert action that automatically gets added to the list of exceed actions if that's set to true (just to avoid needing to copy it for every s6 character)

scenes/game/local_game.gd
Mach131: whoops

scenes/game/game.gd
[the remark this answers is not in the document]
daniel-k-taylor: It should be fine to call _move_card_to_strike_area.
I would kind of expect this to happen automatically as part of _on_strike_started, which looks like is actually happening on your EventType_Strike_RandomGaugeStrike.
I haven't fully analyzed the ordering of all of your events for this random gauge strike thing yet though, so i don't know exactly which events come when.

data/decks/yuzu.json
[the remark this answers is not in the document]
daniel-k-taylor: I think for now we'll just have to pay close attention as we add more character actions that need it.

scenes/game/local_game.gd
daniel-k-taylor: I think I would prefer to have it just all in the character json for clarity/simplicity rather than having to look somewhere else to see what s6_revert_action does.
Then there's less checking of random fields on the character json in the code.
Copying it to each character shouldn't be an issue since when making a new s6 you'll just copy an existing s6 and replace stuff.

The goal is to not have to have like can_do_s6_revert and do_s6_revert in gamewrapper/localgame/remotegame at all.  And in game.gd it just iterates through all character actions and makes buttons for them.

scenes/game/local_game.gd
Mach131: Yeah bonus actions should be fine; I'll also probably still keep the piece of setting active_character_action to false if a strike is happening

daniel-k-taylor: Yes, definitely want to unset this flag to not mess things up later and we can pretty much assume strike will always mean the character action is over.

scenes/menu/main_menu.gd
Mach131: also whoops

data/card_definitions.json
Mach131: As in making it obvious which gauge card will end up at the top of the discard pile? I can try adding in special reminder text and maybe a different highlight color; hesitant to auto-sort gauge in case someone's trying to play around some weird interaction (e.g. mirror of the world)

scenes/game/game.gd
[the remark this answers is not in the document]
Mach131: Oh just to clarify, this particular function is for "Over Here" to try and give a visual of which gauge card was chosen for "X"; I think striking from gauge moves things to the strike area properly, but this effect is a bit weird because the chosen card doesn't ever actually leave gauge

data/card_definitions.json
daniel-k-taylor: Yea, maybe for now just something "Cards discarded in order selected" or something? I dunno. You can try whatever works good in the UI.

scenes/game/game.gd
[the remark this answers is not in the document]
daniel-k-taylor: Oh, I see. In that case, I guess just popping up a notice with the X amount that was chosen would be appropriate?

Mach131:
moved over to https://github.com/daniel-k-taylor/exceedgameclient/pull/4

PR #11, opened 2024-01-16
Ryu

Mach131:
Pretty straightforward; worth noting that if people aren't reading they might accidentally think that the crit prompt when playing an ultra is the validation gauge


data/decks/ryu.json
```diff
@@ -0,0 +1,177 @@
...
+			}
+		},
+	}],
+	"ability_effects": [
+		{
+			"timing": "set_strike",
+			"effect_type": "gauge_for_effect",
+			"per_gauge_effect": null,
+			"overall_effect": {
+				"effect_type": "critical"
```
daniel-k-taylor, line 39: if you set "character_effect": true on this overall_effect effect, will it show up next to the facedown card letting the opponent see more clearly they've used critical?
That's what I did for like Potemkin buff or Gio buff but I don't think those were inside an optional thing.

scenes/game/game.gd
```diff
@@ -2405,6 +2411,8 @@ func can_press_ok():
 				return can_selected_cards_pay_force(select_card_require_force)
 			UISubState.UISubState_SelectCards_GaugeForEffect:
 				if select_card_must_be_max_or_min:
+					if instructions_cancel_allowed and len(selected_cards) == 0:
```
daniel-k-taylor, line 2414: I guess I don't get why this is needed.
How is this different from like Axl Bomber where it's a completely optional thing?

scenes/menu/char_select.tscn
```diff
@@ -293,6 +295,30 @@ layout_mode = 2
...
+[node name="Row1" type="HBoxContainer" parent="CenterContainer/SFCharacterSelect/Rows"]
+layout_mode = 2
+alignment = 1
+
+[node name="Ryu" parent="CenterContainer/SFCharacterSelect/Rows/Row1" instance=ExtResource("6_3lmwy")]
+layout_mode = 2
+char_id = "ryu"
+portrait_texture = ExtResource("30_mommm")
+
+[node name="RandomSF" parent="CenterContainer/SFCharacterSelect/Rows/Row1" instance=ExtResource("6_3lmwy")]
```
daniel-k-taylor, line 317: If you're bored, you can also add a random all seasons button.
I don't know the best way to put that in the UI though.
Someone was requesting it.

daniel-k-taylor (changes requested):
My only real comment is about the character_effect; true thing.
Oh, and adding him to the random ai tests just to make sure the AI can play him without hanging.
Maybe a just ryu test file for the hadoken mechanic would be nice and to give critical some coverage.

scenes/game/game.gd
Mach131: It's something I tried adding just for the UX; without this you can press "OK" without selecting gauge and it does the same thing as pressing "Cancel" (now "Pass"), so I thought it would be clearer to have "OK" only be enabled when it would do something

daniel-k-taylor: ok, that seems fine.  Just double check that like some other per_gauge effect and overall gauge effect works.
If anyone even has a per gauge effect...

data/decks/ryu.json
Mach131: Yup, seems to work nicely

daniel-k-taylor: make sure to put it on both the regular and exceeded versions

scenes/game/game.gd
Mach131: I think they're all on attacks atm (testament, chipp, axl), and I think I've checked all of them

scenes/game/local_game.gd
```diff
@@ -4006,14 +4006,14 @@ func do_card_from_hand_to_gauge(performing_player : Player, card_ids : Array) ->
...
-			
+	
 	var events = []
 	if card_ids.size() > 0:
 		var card_names = card_db.get_card_name(card_ids[0])
 		for i in range(1, card_ids.size()):
 			card_names += ", " + card_db.get_card_name(card_ids[i])
 		var destination_string = ""
-
+
```
daniel-k-taylor, line 4016: I like to use VScode with the option to eliminate extra whitespace when you save so you don't get random stuff like this.
It doesn't matter though, don't worry about removing these.

daniel-k-taylor (approved):

scenes/game/local_game.gd
Mach131: I've just been using the built-in godot editor; i'll try to see if there's anything similar in the settings

PR #18, opened 2024-01-20
Implement Jin

daniel-k-taylor:


scenes/game/local_game.gd
```diff
@@ -1667,11 +1675,23 @@ func advance_to_next_turn():
...
 		events += do_effect_if_condition_met(player_ending_turn, -1, effect, null)
 
+	# Do any end of turn boost effects.
+	# Iterate in reverse as items can be removed.
+	for i in range(len(player_ending_turn.continuous_boosts) - 1, -1, -1):
+		var card = player_ending_turn.continuous_boosts[i]
+		for effect in card.definition['boost']['effects']:
+			if effect['timing'] == "end_of_turn":
+				if card.id in player_ending_turn.end_of_turn_boost_delay_card_ids:
+					# This effect is delayed a turn, so remove it from the list and skip it for now.
```
Mach131, line 1690: I wonder if it'd work nicely to add a field in "end_of_turn" effects to indicate how many turns they go off in, just to avoid needing to have a separate effect to add it to the delay list? Could have a dictionary somewhere to indicate how many turns the boost has skipped to tell when it should go off. I doubt there's anything with a longer waiting time than "end of your next turn" though so might be excessive

scenes/game/local_game.gd
```diff
@@ -3053,13 +3089,33 @@ func handle_strike_effect(card_id :int, effect, performing_player : Player):
...
+
 func get_card_stat(check_player : Player, card : GameCard, stat : String) -> int:
 	var value = card.definition[stat]
 	if str(value) == "X":
 		if active_strike:
 			return check_player.strike_stat_boosts.strike_x
 		else:
 			assert(false, "ERROR: No support for interpreting X outside of strikes")
+	elif str(value) == "CARDS_IN_HAND":
+		value = check_player.hand.size()
```
Mach131, line 3118: There's nothing that happens to print out cards' coded guard values or anything right? I was worrying about how to do things like merkava's distance from opponent's Xs but this is probably an easier solution than what I was thinking of

Mach131 (approved):

scenes/game/local_game.gd
daniel-k-taylor: This list could be a dictionary, I just didn't think there was a need for it.
I kind of wanted to track the turns remaining in the effect itself, but that just felt kind of nasty.
I think I'll just stick with this.

scenes/game/local_game.gd
daniel-k-taylor: There isn't. And if there is, it needs to account for these things.

PR #19, opened 2024-01-20
Bug fixes for block, septem, mid strike boosts, evaluating effect conditions later to allow enabling them mid strike

daniel-k-taylor:
* Always ask for block
* Septem Voices effects not orderable
* Jacko throw servant after effects happen immediately
* Condition effects evaluated later (Nu13 Exceed Sword of destruction)

### DIFF
--- a/scenes/game/local_game.gd
+++ b/scenes/game/local_game.gd
@@ -1570,7 +1570,8 @@ class Player:
 		var effects = []
 
 		# Maybe later get them from boosts, but for now, just character ability.
-		effects = parent.get_all_effects_for_timing("set_strike", self, card)
+		var ignore_condition = false
+		effects = parent.get_all_effects_for_timing("set_strike", self, card, ignore_condition)
 
 		if extra_effect_after_set_strike:
 			effects.append(extra_effect_after_set_strike)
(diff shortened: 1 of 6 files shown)


scenes/game/local_game.gd
```diff
@@ -3199,18 +3200,40 @@ func do_remaining_effects(performing_player : Player, next_state):
...
-			change_game_state(Enums.GameState.GameState_PlayerDecision)
-			decision_info.type = Enums.DecisionType.DecisionType_ChooseSimultaneousEffect
-			decision_info.player = performing_player.my_id
-			decision_info.choice = active_strike.remaining_effect_list
-			events += [create_event(Enums.EventType.EventType_Strike_EffectChoice, performing_player.my_id, 0, "EffectOrder")]
-			break
+			# Check to see if any of these effects actually have their condition met (or have a negative condition).
+			# If more than 1, send only those choices to the player.
+			# If only 1 does, remove it from the list and do it immediately.
+			# If none do, this is over, clear out the list.
```
Mach131, line 3206: Even if we decide we want to show the player effects with unmet conditions, maybe it could be nice to repurpose this to indicate that somehow (e.g. changing the button color?) Clearing the list when nothing is met is probably safe either way though yeah

Mach131 (approved):

scenes/game/local_game.gd
daniel-k-taylor: Agreed. This will still be useful later.

PR #23, opened 2024-01-21
Plague knight

Mach131:
Not bad apart from some strange boosts, though the UI on his UA can be weird


data/card_definitions.json
```diff
@@ -6443,6 +6443,361 @@
...
+		],
+		"boost": {
+			"boost_type": "immediate",
+			"force_cost": 1,
+			"cancel_cost": -1,
+			"display_name": "Bait Bomb",
+			"effects": [
+				{
+					"timing": "immediate",
+					"effect_type": "choose_cards_from_top_deck",
```
daniel-k-taylor, line 6724: Maybe I'm oversimplifying, but isn't this boost actually:
Look at the top 3 cards, add 1 to gauge, draw the other 2 (1 for effect 1 for end of turn)?

scenes/game/game.gd
```diff
@@ -2374,7 +2384,18 @@ func _update_buttons():
...
+		if len(card_text) > ChoiceTextLengthSoftCap:
+			var break_idx = ChoiceTextLengthSoftCap-1
+			while break_idx < len(card_text)-1 and card_text[break_idx] != " ":
+				break_idx += 1
+				if break_idx >= ChoiceTextLengthHardCap:
+					break
+			if card_text[break_idx] == " ":
+				card_text = card_text.substr(0, break_idx) + "\n" + card_text.substr(break_idx+1)
+			else:
+				card_text = card_text.substr(0, break_idx) + "-\n" + card_text.substr(break_idx)
```
daniel-k-taylor, line 2397: Does this still make sense if the last word of the text spanned the soft cap limit?
In that case, break_idx should be = len(card_text)-1, which is the last character.

scenes/game/local_game.gd
```diff
@@ -2382,6 +2405,35 @@ func handle_strike_effect(card_id :int, effect, performing_player : Player):
 			var close_amount = abs(performing_start - new_location)
 			local_conditions.fully_closed = close_amount == effect['amount']
 			_append_log("%s Close %s - Moved from %s to %s." % [performing_player.name, str(effect['amount']), str(previous_location), str(new_location)])
+		"copy_other_hit_effect":
+			var card = null
+			if performing_player == active_strike.initiator:
```
daniel-k-taylor, line 2410: You should use active_strike.get_player_card(performing_player : Player) -> GameCard

scenes/game/local_game.gd
```diff
@@ -2382,6 +2405,35 @@ func handle_strike_effect(card_id :int, effect, performing_player : Player):
...
+				card = active_strike.initiator_card
+			else:
+				card = active_strike.defender_card
+			var hit_effects = get_all_effects_for_timing("hit", performing_player, card)
+
+			var effect_options = []
+			for possible_effect in hit_effects:
+				if possible_effect != effect:
```
daniel-k-taylor, line 2418: Do you need to have something in here to exclude the other copy of this boost?
I expect possible_effect != effect is doing a reference equality comparison check and not a deep comparison.

scenes/game/local_game.gd
```diff
@@ -3134,6 +3200,36 @@ func get_striking_card_ids_for_player(check_player : Player) -> Array:
...
+					as_character_effect['character_effect'] = true
+					card_effects.append(as_character_effect)
+				else:
+					card_effects.append(card_effect)
+		new_effects += card_effects
+	performing_player.strike_stat_boosts.added_attack_effects += new_effects
+
+func duplicate_attack_triggers(performing_player : Player, amount : int):
+	var card = null
+	if active_strike.initiator == performing_player:
```
daniel-k-taylor, line 3222: use active_strike.get_player_card

scenes/game/local_game.gd
```diff
@@ -3216,7 +3318,17 @@ func do_remaining_effects(performing_player : Player, next_state):
 				change_game_state(Enums.GameState.GameState_PlayerDecision)
 				decision_info.type = Enums.DecisionType.DecisionType_ChooseSimultaneousEffect
 				decision_info.player = performing_player.my_id
-				decision_info.choice = effects_to_choose
+				decision_info.choice = []
+				for effect in effects_to_choose:
+					if 'is_negative_effect' in effect and effect['is_negative_effect']:
```
daniel-k-taylor, line 3323: This was duplicated somewhere else, good candidate for a function?
It might not be exactly the same, but at least consider if it makes sense.

scenes/game/local_game.gd
```diff
@@ -3410,13 +3525,13 @@ func in_range(attacking_player, defending_player, card):
...
 			return false
-	if min_range <= distance and distance <= max_range:
-		return true
-	return false
+	return opponent_in_range
 
 func calculate_damage(offense_player : Player, defense_player : Player, offense_card : GameCard, defense_card : GameCard) -> int:
 	var power = get_card_stat(offense_player, offense_card, 'power')
 	var damage = power + offense_player.strike_stat_boosts.power
+	if power == -1:
```
daniel-k-taylor, line 3533: Rather than do this, maybe we should just add this to block:
"effect_type": "attack_does_not_hit"
Makes it a lot more explicit than this fragile -1 power thing.

daniel-k-taylor: If we really want the power to be right, maybe add a "total_power_is_always_X" effect, I think that'll come in handy for Orie too.

daniel-k-taylor (changes requested):

data/card_definitions.json
Mach131: I should check if there's any order enforced on it, but the situation i'm concerned about is that if there's one card left in deck they should be able to choose whether it goes to Gauge or Hand

daniel-k-taylor: That's annoying.

scenes/game/local_game.gd
Mach131: Changed it to check the effect type instead; hopefully safe to assume this effect won't show up in an "and" or anything

scenes/game/local_game.gd
Mach131: not hitting should be taken care of by the range fix; i'll try adding the effect just to be safe

daniel-k-taylor: The main reason I also don't want you to do this here is that I changed this with the Phonon change, so you'll have to merge some conflicts.
I think it's much better to not have -1 be a special power number.
Similarly, if block had the attack_does_not_hit effect, then the range would also be irrelevant.

daniel-k-taylor (dismissed):

scenes/game/local_game.gd
Mach131: ooh I see; the phonon conflicts should all be merged now and I borrowed the idea of overwriting power, so should be fine on the current version (though I can throw on attack_does_not_hit quickly)

daniel-k-taylor (approved):

PR #32, opened 2024-01-25
Implement Noel

Rose-Personal:


data/card_definitions.json
```diff
@@ -6485,6 +6485,383 @@
...
+				"effect_type": "stun_immunity"
+			},
+			{
+				"timing": "before",
+				"condition": "was_hit",
+				"effect_type": "powerup",
+				"amount": 4
+			},
+			{
+				"timing": "before",
```
Mach131, line 6513: It may be preferable to make this effect an "and" field under the above; that way it won't need to ask the player which one they want to resolve first (plus you won't need to repeat the timing and condition)

data/card_definitions.json
```diff
@@ -6485,6 +6485,383 @@
...
+					"amount2": 2
+				},
+				"force_max": 1
+			},
+			{
+				"timing": "after",
+				"effect_type": "choice",
+				"choice": [
+					{ "effect_type": "pass" },
+					{ "effect_type": "advance", "amount": 2 }
```
Mach131, line 6747: I think we usually put pass as the last choice, just for consistency

data/decks/noel.json
```diff
@@ -0,0 +1,179 @@
...
+			{ "effect_type": "close", "amount": 1 },
+			{ "effect_type": "retreat", "amount": 1 }
+		]
+	},
+	"ability_effects": [
+		{
+			"timing": "on_reshuffle",
+			"condition": "manual_reshuffle",
+			"effect_type": "add_set_aside_card_to_deck",
+			"id": "noel_valkyrieveil"
```
Mach131, line 23: make sure to add this as an exceed ability effect as well (we forgot in the existing s5 files, should be fixed soon)

data/decks/noel.json
```diff
@@ -0,0 +1,179 @@
...
+		},
+		{
+			"timing": "set_strike",
+			"condition": "initiated_strike",
+			"effect_type": "gauge_for_effect",
+			"per_gauge_effect": null,
+			"overall_effect": {
+                "character_effect": true,
```
Mach131, line 31: probably doesn't matter a huge amount, but it may be nice to line up the indentation here and in the exceed ability effects (unless that's just a github formatting thing?)

Mach131 (changes requested):
Mostly looks good, just a few minor formatting things mostly~ Also, if you want, you can add a text effect description for "look_at_opponent_hand" in globals/card_definitions.gd; it probably doesn't matter since it's only used as a now effect, but might be nice for futureproofing

data/decks/noel.json
Rose-Personal: think this one is a github formatting thing, not seeing it in my own source files

Mach131 (approved):
looks good to me! will probably wait until DT sees this to merge (do you already have merge permissions? i'm not sure if i can give them myself)

Rose-Personal:
I do not, sounds like a reasonable course of action to me!

data/card_definitions.json
```diff
@@ -6485,6 +6485,381 @@
...
+		],
+		"boost": {
+			"boost_type": "immediate",
+			"force_cost": 0,
+			"cancel_cost": -1,
+			"display_name": "Power of the Eye",
+			"effects": [
+				{
+					"timing": "immediate",
+					"effect_type": "look_at_opponent_hand"
```
daniel-k-taylor, line 6571: I know this is kind of annoying, but can you use the "reveal_hand" effect we already have an add an extra field that is "opponent": true?
Then just have an if statement in the "reveal_hand" handling code that says:
if 'opponent' in effect and effect['opponent']:
    opposing_player.reveal_hand()
else:
    performing_player.reveal_hand()

data/decks/noel.json
```diff
@@ -0,0 +1,185 @@
...
```
daniel-k-taylor, line 31: This is a tabs vs spaces thing.
Can you make it tabs to be consistent with the rest?

daniel-k-taylor (changes requested):

daniel-k-taylor (approved):

PR #34, opened 2024-01-25
Reference on reshuffle improved

Mach131:
Reshuffling reveals remaining cards, brings up the reference UI for the opponent to acknowledge, and Hopefully Does Not Break Any Draw/Wild Swing Effects

### DIFF
--- a/scenes/game/local_game.gd
+++ b/scenes/game/local_game.gd
@@ -918,6 +918,7 @@ class Player:
 				deck.remove_at(0)
 				events += [parent.create_event(Enums.EventType.EventType_Draw, my_id, card.id)]
 			else:
+				parent.printlog("WARNING: probably shouldn't need to call reshuffle from draw anymore")
 				events += reshuffle_discard(false)
 				if not parent.game_over:
 					var card = deck[0]
@@ -933,7 +934,7 @@ class Player:
 			deck.insert(0, card)
 		return events
 
-	func begin_reshuffle(manual : bool):
+	func begin_reshuffle(manual : bool, followup_effect = null):
 		var events : Array = []
 		if reshuffle_remaining == 0:
 			# Game Over
@@ -942,9 +943,11 @@ class Player:
 		else:
 			# Prompt other player to review discards
 			parent.change_game_state(Enums.GameState.GameState_PlayerDecision)
+			parent.decision_info.player = my_id
 			parent.decision_info.type = Enums.DecisionType.DecisionType_ReviewReshuffle
 			parent.decision_info.source = manual
-			events += [parent.create_event(Enums.EventType.EventType_BeginReshuffle, my_id, 0, "", manual)]
+			parent.decision_info.bonus_effect = followup_effect
+			events += [parent.create_event(Enums.EventType.EventType_BeginReshuffle, my_id, 0, "", manual, followup_effect)]
 		return events
 
 	func reshuffle_discard(manual : bool):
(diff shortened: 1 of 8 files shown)


scenes/game/local_game.gd
```diff
@@ -933,6 +935,36 @@ class Player:
 			deck.insert(0, card)
 		return events
 
+	func get_unknown_cards():
+		var unknown_cards = hand + deck # TODO: secret sealed areas
```
daniel-k-taylor, line 939: don't forget this

scenes/game/local_game.gd
```diff
@@ -933,6 +935,36 @@ class Player:
...
+		parent.shuffle_array(unknown_cards)
```
daniel-k-taylor, line 940: This is going to screw things up, you're using the random seed only on one client since you're calling this from game_wrapper but only for one player, right?

scenes/game/local_game.gd
```diff
@@ -918,6 +919,7 @@ class Player:
 				deck.remove_at(0)
 				events += [parent.create_event(Enums.EventType.EventType_Draw, my_id, card.id)]
 			else:
+				parent.printlog("WARNING: probably shouldn't need to call reshuffle from draw anymore")
```
daniel-k-taylor, line 922: What is this?

scenes/game/local_game.gd
```diff
@@ -2593,7 +2629,20 @@ func handle_strike_effect(card_id :int, effect, performing_player : Player):
...
+			else:
+				# Draw what you can now
+				var remainder = draw_amount - deck_size
+				events += performing_player.draw(deck_size)
+
+				# Process reshuffle, and prepare the remainder of the draw for after
+				var followup_effect = effect.duplicate()
+				followup_effect['card_id'] = card_id
+				followup_effect['amount'] = remainder
+				events += performing_player.begin_reshuffle(false, followup_effect)
```
daniel-k-taylor, line 2645: I really, really don't like adding a decision inside draw, as basically in all cards created so far, the effects were written on the assumption that draw is not a decision point.
I also don't like moving the reshuffling logic outside of player.draw.
I don't want to have a blocking operation on the current player's turn for something as simple as drawing.

Instead of doing it this way, can we just send an event on reshuffle that provides the list of unknown cards somehow that then updates revealed cards?
Additionally, you could add something that marks up the deck reference somehow.

I'm just really worried this will break some obscure cases and make it more complicated to write all effects knowing draw is now a decision point.

daniel-k-taylor:
I don't think I like this change.
I would much rather draw remain a non-decision point and instead just make it easier for the opponent to bring up some UI to see the reshuffle.
Maybe it should be brought up automatically for them by the game scene the next time it is their turn to make a decision just as a reminder?

daniel-k-taylor (changes requested):

scenes/game/local_game.gd
Mach131: This is mostly a reminder for when we implement a character that has a secret sealed area (specter/arakune)

scenes/game/local_game.gd
Mach131: oh good point; i was thinking of just sorting by ID instead so that's probably a good reason to do so

scenes/game/local_game.gd
Mach131: oh I forgot to remove this; i was making sure nothing hit this part while testing

scenes/game/local_game.gd
Mach131: As mentioned in DMs it's probably better to go with an approach like this; I do think this branch is functional so I'll keep it around personally, but I'll probably close this PR and open a new one once I have that version done

PR #36, opened 2024-01-26
Hazama

daniel-k-taylor:
Implement Hazama
Bug fix for Nago cleanup (should be normals only)
Put checking in the animation spriteframes so we don't get spurious errors in the console.


scenes/game/character.gd
```diff
@@ -74,19 +80,29 @@ func set_exceed(is_exceed : bool):
...
 
+func set_buddy_id(id : String):
+	buddy_id = id
+
+func get_buddy_id():
+	return buddy_id
+
+func play_animation(named_animation : String):
+	if animation.sprite_frames.has_animation(named_animation):
+		animation.play(named_animation)
```
Mach131, line 91: not at all important, but it would be nice if something here could have have run/dash_back play the walking animations instead if they're not available

Mach131 (approved):
A lot going on, but seems reasonable to me given that everything's working

PR #42, opened 2024-01-27
Implement Guile

daniel-k-taylor:


scenes/game/local_game.gd
```diff
@@ -5318,6 +5370,10 @@ func do_choice(performing_player : Player, choice_index : int) -> bool:
 				active_exceed = false
 				if game_state != Enums.GameState.GameState_WaitForStrike:
 					events += check_hand_size_advance_turn(performing_player)
+			elif active_change_cards:
+				active_change_cards = false
```
Mach131, line 5374: It hits this line regardless of which option is picked right? I would be worried it doesn't get turned off if you strike but it seems like it shouldn't set active_strike until the next event is handled, so hopefully fine

Mach131 (approved):
Seems good to me, though unfortunately specific

scenes/game/local_game.gd
daniel-k-taylor: yes

PR #45, opened 2024-01-28
Implemented Linne

Rose-Personal:
Needs testing but please let me know if implementation changes should be made first


data/card_definitions.json
```diff
@@ -5813,6 +5813,367 @@
...
+					}
+				},
+				{
+					"timing": "now",
+					"effect_type": "strike"
+				}
+			]
+		}
+	},
+
```
daniel-k-taylor, line 5866: Can you only leave extra newlines between cards when it is different characters?

data/decks/linne.json
```diff
@@ -0,0 +1,190 @@
...
+					]
+				}
+			}
+		}
+	}],
+	"character_action_exceeded": [
+		{
+			"gauge_cost": 0,
+			"force_cost": 0,
+			"can_continuous_boost_from_gauge": true,
```
daniel-k-taylor, line 31: maybe "requires_continuous_boost_in_gauge" would be more descriptive?

data/decks/linne.json
```diff
@@ -0,0 +1,190 @@
...
+			"effect_type": "powerup",
+			"amount": 1
+		}
+	],
+	"cards": [
+		{
+			"definition_id": "linne_divineblaze",
+			"image": "linne_divineblaze.tres"
+		},
+        {
```
daniel-k-taylor, line 73: tabs v spaces?

scenes/game/card_database.gd
```diff
@@ -42,6 +42,12 @@ func get_card_name(id : int) -> String:
 			return card.definition['display_name']
 	return "MISSING CARD"
 
+func get_card_name_by_card_id(id : String) -> String:
```
daniel-k-taylor, line 45: maybe call this by_card_definition_id?

scenes/game/game.gd
```diff
@@ -1741,7 +1743,13 @@ func begin_strike_choosing(strike_response : bool, cancel_allowed : bool,
 	select_card_require_min = 1
 	select_card_require_max = 1
 	var can_cancel = cancel_allowed and not strike_response
-	enable_instructions_ui("Select a card to strike with.", true, can_cancel, not disable_wild_swing, not disable_ex)
+	var dialogue = "Select a card to strike with."
+	var player = game_wrapper._get_player(Enums.PlayerId.PlayerId_Player)
```
daniel-k-taylor, line 1747: so, I pretty explicitly tried to not directly call things internal to LocalGame/Player from Game as an effort to keep the UI and game logic separate.

Can you add a function to game_wrapper for what you're trying to check?
Something like game_wrapper.get_sidestep_card_name() or something? and then add it to dialogue if it is not empty?

scenes/game/local_game.gd
```diff
@@ -85,6 +85,7 @@ func create_event(event_type : Enums.EventType, event_player : Enums.PlayerId, n
 		"reason": reason,
 		"extra_info": extra_info,
 		"extra_info2": extra_info2,
+		"extra_info3": extra_info3,
```
daniel-k-taylor, line 88: I really should have made this more variable, lol...
This is fine though, just leave it.
Surely we won't need any more than 3 extra things, right?

scenes/game/local_game.gd
```diff
@@ -407,6 +409,7 @@ class Player:
 	var movement_limit : int
 	var free_force : int
 	var guile_change_cards_bonus : bool
+	var will_not_hit : Array[String]
```
daniel-k-taylor, line 412: let's go with cards_that_will_not_hit.

scenes/game/local_game.gd
```diff
@@ -1386,6 +1395,7 @@ class Player:
 		arena_location = new_arena_location
 		events += [parent.create_event(Enums.EventType.EventType_Move, my_id, new_arena_location, "move", distance, previous_location)]
 		if position_changed:
+			moved_self_this_strike = true
```
daniel-k-taylor, line 1398: Would this not go within on_position_changed?
Maybe just add a param for is_self_move if the point is not to include push/pull, maybe it'll be useful for something else later.

scenes/game/local_game.gd
```diff
@@ -2531,6 +2550,11 @@ func handle_strike_effect(card_id :int, effect, performing_player : Player):
 		"armorup":
 			performing_player.strike_stat_boosts.armor += effect['amount']
 			events += [create_event(Enums.EventType.EventType_Strike_ArmorUp, performing_player.my_id, effect['amount'])]
+		"armorup_damage_dealt":
+			## This is inaccurate if you have this effect on an additional attack character -- currently impossible, and additional attacks mess up many things
```
daniel-k-taylor, line 2554: Just one # for comments.
Can you explain exactly what is inaccurate?
I guess we haven't implemented additional attacks, but I would expect get_damage_taken to return the total damage of all multiple strikes?

scenes/game/local_game.gd
```diff
@@ -3542,6 +3586,22 @@ func handle_strike_effect(card_id :int, effect, performing_player : Player):
...
+		"sidestep_transparent_foe":
+			change_game_state(Enums.GameState.GameState_PlayerDecision)
+			decision_info.type = Enums.DecisionType.DecisionType_Sidestep
+			decision_info.effect_type = "sidestep_internal"
+			decision_info.choice_card_id = card_id
+			decision_info.player = performing_player.my_id
+			events += [create_event(Enums.EventType.EventType_Boost_Sidestep, performing_player.my_id, 0)]
+		"sidestep_dialogue":
+			pass
+			# this exists purely for ui, no-op here
```
daniel-k-taylor, line 3598: put this comment above the pass

scenes/game/local_game.gd
```diff
@@ -4358,12 +4425,14 @@ func continue_resolve_strike(events):
...
 					_append_log("Range Check: %s (%s) vs %s (%s)." % [player1.name, player1.arena_location, player2.name, player2.arena_location])
-				if in_range(player1, player2, card1):
+				if in_range(player1, player2, card1) and not card1.definition['id'] in player1.will_not_hit:
 					_append_log("%s %s hits." % [player1.name, card_db.get_card_name(card1.id)])
 					active_strike.player1_hit = true
 					active_strike.strike_state = StrikeState.StrikeState_Card1_Hit
 					active_strike.remaining_effect_list = get_all_effects_for_timing("hit", player1, card1)
 				else:
+					if card1.definition['id'] in player1.will_not_hit:
+						_append_log("Named card %s does not hit this strike!" % card_db.get_card_name(card1.id))
```
daniel-k-taylor, line 4435: Maybe use an else here and don't print the misses line if you print this one?
Actually, maybe combine them so it says %s %s misses (this was the named card)?
i would do this with like:
var extra_details = ""
if ....
  extra_details = " (this was the named card)"
 append_log("%s %s misses%s." % [player1.name, card_db.get_card_name(card1.id), extra_details])

scenes/game/local_game.gd
```diff
@@ -4397,12 +4466,14 @@ func continue_resolve_strike(events):
...
 					_append_log("Range Check: %s (%s) vs %s (%s)." % [player2.name, player2.arena_location, player1.name, player1.arena_location])
-				if in_range(player2, player1, card2):
+				if in_range(player2, player1, card2) and not card2.definition['id'] in player2.will_not_hit:
 					_append_log("%s %s hits." % [player2.name, card_db.get_card_name(card2.id)])
 					active_strike.player2_hit = true
 					active_strike.strike_state = StrikeState.StrikeState_Card2_Hit
 					active_strike.remaining_effect_list = get_all_effects_for_timing("hit", player2, card2)
 				else:
+					if card2.definition['id'] in player2.will_not_hit:
+						_append_log("Named card %s does not hit this strike!" % card_db.get_card_name(card2.id))
```
daniel-k-taylor, line 4476: same as player1 case, see above

daniel-k-taylor (changes requested):
Looks pretty good, just some minor suggestions for naming and tweaking a few things.

data/decks/linne.json
Rose-Personal: I don't _think_ this distinction matters with the card pool we have, but if Linne or Wagner had boosts with force costs and no ability to pay them that would also prevent this action. Perhaps something like "can_play_continuous_boost_from_gauge"? Given I think the difference is academic, don't feel strongly if you prefer requires.

data/decks/linne.json
Rose-Personal: Sorry, not sure why my editor is doing this...

scenes/game/local_game.gd
Rose-Personal: The correct logic for this card only gets armor for the damage that it itself does, but if this were played as an additional attack it would get incorrect amounts of armor since get_damage_taken works the way you're describing. Will try to make comment clearer, don't imagine this situation will ever be possible even in a new set.

Rose-Personal:
Made suggested changes. In the case of the character ability, currently using "can_boost_continuous_boost_from_gauge", but can still change that if requires is better. Otherwise, tests and then ready for real PR.

data/decks/linne.json
daniel-k-taylor: yea, let's go with "can_play" then

scenes/game/game_wrapper.gd
```diff
@@ -207,6 +207,15 @@ func get_player_force_for_cards(player_id : Enums.PlayerId, card_ids : Array, re
 func get_force_to_move_to(player_id : Enums.PlayerId, location : int):
 	return _get_player(player_id).get_force_to_move_to(location)
 
+func get_will_not_hit_card_names(player_id : Enums.PlayerId) -> Array:
+	var card_names = []
+	var player = _get_player(player_id)
+	if player.cards_that_will_not_hit.size() > 0:
+		var card_db = get_card_database()
+		for card in player.cards_that_will_not_hit:
+			card_names += [card_db.get_card_name_by_card_definition_id(card)]
```
daniel-k-taylor, line 216: This works, it's just a little weird to use array += instead of just .append() when adding a single item.

scenes/game/local_game.gd
```diff
@@ -3594,14 +3592,14 @@ func handle_strike_effect(card_id :int, effect, performing_player : Player):
...
 			# this exists purely for ui, no-op here
+			pass
 		"sidestep_internal":
 			var named_card = card_db.get_card(effect['card_id'])
 			# named_card is the individual card but
-			# this should discard "by name", so instead of using that
+			# this should match "by name", so instead of using that
 			# match card.definition['id']'s instead.
-			opposing_player.will_not_hit += [named_card.definition['id']]
+			opposing_player.cards_that_will_not_hit += [named_card.definition['id']]
```
daniel-k-taylor, line 3602: again, kind of weird to not just use .append()

daniel-k-taylor (changes requested):
Looks good. Just a comment on using array append instead of array += in a couple places.
I only used += for events all the time because I didn't want to get mixed up by having append and += and using the wrong one.

I'd like to see at least a few tests for Linne around the new mechanics (armor per damage, her special boost thing, per boosts in play, sidestep).

data/decks/linne.json
daniel-k-taylor: actually, what you have now is fine

daniel-k-taylor (approved):

PR #46, opened 2024-01-28
Sort the player hand

daniel-k-taylor:


Mach131 (commented):
Seems reasonable to me; would it be possible to pull out some of the GameCard level sorting logic to a separate function (maybe in card_db)? might be nice to steal for things like the reshuffle reveal ordering

PR #47, opened 2024-01-28
Sagat

daniel-k-taylor:


data/card_definitions.json
```diff
@@ -11368,6 +11368,362 @@
...
+		}
+	},
+	{
+		"id": "sagat_lowstepkick",
+		"type": "special",
+		"display_name": "Low Step Kick",
+		"force_cost": 0,
+		"gauge_cost": 0,
+		"range_min": 1,
+		"range_max": 1,
```
Mach131, line 11514: Max range 2 I think

Mach131 (changes requested):
Looks good I think, just one stat error

Mach131 (approved):

data/decks/sagat.json
```diff
@@ -0,0 +1,189 @@
...
+				"character_effect": true
+			},
+			"gauge_max": 1
+		},
+		{
+			"timing": "set_strike",
+			"condition": "initiated_strike",
+			"effect_type": "choice",
+			"choice": [
+				{ "effect_type": "reveal_strike", "and": { "effect_type": "speedup", "amount": 1 } },
```
Mach131, line 36: can you put the character_effect = true thing together with the speed boost, along with the speed + power boost in the exceed mode? that just makes the reminder popup happen which is helpful

Mach131 (changes requested):

PR #48, opened 2024-01-29
Fancy logging

Mach131:
Making a filterable logging system; probably needs a lot of refinement so I kept all of the old logging stuff in place

### DIFF
--- a/scenes/game/game.gd
+++ b/scenes/game/game.gd
@@ -525,7 +525,8 @@ func _process(delta):
 		var events = game_wrapper.poll_for_events()
 		if events.size() > 0:
 			_handle_events(events)
-			combat_log.set_text(game_wrapper.get_combat_log(combat_log.get_filters()))
+			var log_text = game_wrapper.get_combat_log(combat_log.get_filters(), combat_log.log_player_color, combat_log.log_opponent_color)
+			combat_log.set_text(log_text)
 		elif ui_state == UIState.UIState_WaitingOnOpponent:
 			# Advance the AI game automatically.
 			_on_ai_move_button_pressed()
@@ -3647,7 +3648,8 @@ func _on_card_popout_pressed_toggle():
 
 
 func _on_combat_log_button_pressed():
-	combat_log.set_text(game_wrapper.get_combat_log(combat_log.get_filters()))
+	var log_text = game_wrapper.get_combat_log(combat_log.get_filters(), combat_log.log_player_color, combat_log.log_opponent_color)
+	combat_log.set_text(log_text)
 	combat_log.visible = true
 
 func _on_combat_log_close_button_pressed():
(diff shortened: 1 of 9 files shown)


scenes/game/game.gd
```diff
@@ -523,7 +523,10 @@ func _process(delta):
 		var events = game_wrapper.poll_for_events()
 		if events.size() > 0:
 			_handle_events(events)
-			$CombatLog.set_text(game_wrapper.get_combat_log())
+			if $CombatLog.use_old_log:
```
daniel-k-taylor, line 526: can you go ahead and define this with an @onready var at the top and use that instead of $CombatLog?

scenes/game/local_game.gd
```diff
@@ -18,6 +18,9 @@ const MinArenaLocation = 1
 const MaxArenaLocation = 9
 const ShuffleEnabled = true
 
+const log_player_color = "red"
```
daniel-k-taylor, line 21: feels like a UI thing in game.gd rather than in the local_game logic
probably fine for now i guess, there's not a great split between UI/game logic for the log
but this is something I would expect to see defined in combat_log.gd or something

scenes/game/local_game.gd
```diff
@@ -674,6 +728,11 @@ class Player:
 
 	func shuffle_sealed_to_deck():
 		var events = []
+		var card_names = parent._card_list_to_string(sealed)
+		if card_names != "":
```
daniel-k-taylor, line 732: this can just be if card_names:

scenes/game/local_game.gd
```diff
@@ -5481,7 +5904,7 @@ func do_card_from_hand_to_gauge(performing_player : Player, card_ids : Array) ->
 	return true
 
 func do_boost_name_card_choice_effect(performing_player : Player, card_id : int) -> bool:
-	var card_name = card_db.get_card_name(card_id)
+	var card_name = "the card " + card_db.get_card_name(card_id)
 	if card_id == -1:
 		card_name = "a nonexistent card"
 	printlog("SubAction: BOOST_NAME_CARD by %s card %s" % [get_player_name(performing_player.my_id), card_name])
```
daniel-k-taylor, line 5910: by PLAYER card the card CARDNAME?
Gonna make the printlog look weird
though this is just for debug

daniel-k-taylor (changes requested):
I looked through a decent chunk. I guess I don't get why you didn't just completely replace the existing log. Having 2 different ways to log stuff just gets confusing.

scenes/game/local_game.gd
Mach131: probably a good idea; I can look into making it adjustable as well

scenes/game/local_game.gd
Mach131: just changed it to "named %s" to help a bit

daniel-k-taylor (approved):

PR #50, opened 2024-01-30
Change end of strike discard order, minor fixes

Rose-Personal:
Shitsu after effect now resolves in one prompt
Removed extraneous prints from a Noel test case
Carl correctly handles opponent flipping nirvana at range 0

### DIFF
--- a/scenes/game/local_game.gd
+++ b/scenes/game/local_game.gd
@@ -4849,7 +4849,7 @@ func continue_resolve_strike(events):
 			StrikeState.StrikeState_Cleanup_Player2Effects:
 				events += do_remaining_effects(player2, StrikeState.StrikeState_Cleanup_Complete)
 			StrikeState.StrikeState_Cleanup_Complete:
-				# If hit, move card to gauge, otherwise move to discard.
+				# Handle cleanup effects that cause attack cards to leave play before the standard timing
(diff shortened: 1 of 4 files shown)


scenes/game/local_game.gd
```diff
@@ -4859,6 +4861,12 @@ func continue_resolve_strike(events):
 				events += player1.cleanup_continuous_boosts()
 				events += player2.cleanup_continuous_boosts()
 
+				# If hit, move card to gauge, otherwise move to discard.
+				if card1 in active_strike.cards_in_play:
+					events += strike_send_attack_to_discard_or_gauge(player1, card1)
+				if card2 in active_strike.cards_in_play:
+					events += strike_send_attack_to_discard_or_gauge(player2, card2)
+
```
daniel-k-taylor, line 4869: want to just put in an assert after this that active_strike.cards_in_play.size() == 0?

Can you also add comments above to explain that handle_strike_attack_cleanup() handles all effects aside from the standard hit/miss and that those are handled down here because boosts go to discard first?

daniel-k-taylor (changes requested):
Looks good, would just prefer some extra comments explaining why things are where they are and maybe add an assert.

daniel-k-taylor (approved):

PR #55, opened 2024-01-31
Implementing Mika

Mach131:
UA a bit weird, but not bad otherwise


data/card_definitions.json
```diff
@@ -7409,6 +7409,340 @@
...
+			"boost_type": "immediate",
+			"force_cost": 0,
+			"cancel_cost": -1,
+			"display_name": "Smart Steer",
+			"effects": [
+				{
+					"timing": "immediate",
+					"effect_type": "choice",
+					"choice": [
+						{ "effect_type": "advance", "amount": 1 },
```
daniel-k-taylor, line 7501: maybe it is time to change these into like "move_to_any_space" but with a "move_min": 0, "move_max": 3 or something.
That'll bring up the space selection UI instead of the choice box and probably be a better overall experience.

daniel-k-taylor: Figuring out exactly which spaces to allow will be a little annoying since you have to consider retreat into the edge and account for the opponent, but I think it'll be better in the end.

data/decks/mika.json
```diff
@@ -0,0 +1,182 @@
+{
+	"id": "mika",
+	"season": 6,
+	"display_name": "Mika Returna",
+	"cardback": "cardback_mika.jpg",
+	"exceed_cost": 1,
+	"character_action_default": [{
+		"gauge_cost": 0,
+		"force_cost": 0,
+		"effect": {
```
daniel-k-taylor, line 10: should this have requires min cards in hand 1?

data/decks/mika.json
```diff
@@ -0,0 +1,182 @@
...
+			"effect_type": "self_discard_choose",
+			"optional": false,
+			"amount": 1,
+			"and": {
```
daniel-k-taylor, line 14: Does and work after this?
maybe use the discard_effect stuff I added.

scenes/game/game.gd
```diff
@@ -90,6 +90,7 @@ var cached_opponent_location = 0
 var reference_popout_toggle_enabled = false
 var reference_popout_toggle = false
 var opponent_cards_before_reshuffle = []
+var boost_selection_saved = {}
```
daniel-k-taylor, line 93: I feel like there has to be a better name.
Maybe boost_selection_options or something.

daniel-k-taylor (changes requested):
Just some minor comments

data/decks/mika.json
Mach131: the "and" actually is supported by the character action handler, and I wanted to avoid discard_effect because of discarding 0 technically being legal (though as mentioned above we might not have to worry about that)

scenes/game/local_game.gd
```diff
@@ -3518,16 +3518,37 @@ func handle_strike_effect(card_id :int, effect, performing_player : Player):
...
+			if 'move_min' in effect:
+				move_min = effect['move_min']
+			if 'move_max' in effect:
+				move_max = effect['move_max']
+
+			decision_info.limitation = []
+			if move_min == 0:
+				decision_info.limitation.append(0)
+				decision_info.choice.append({
+					"effect_type": "move_to_space",
```
daniel-k-taylor: this also needs to be added if say you're at location 9 and move_min and move_max > 1, to account for retreat 1 into corner.

daniel-k-taylor: also, we should double check that this works with required movement, move_min>1 and movement is blocked.

daniel-k-taylor (changes requested):

scenes/game/local_game.gd
Mach131: should be improved after the latest push; I held off on applying it to existing movement effects but I can try to if it looks good

scenes/game/local_game.gd
```diff
@@ -3508,44 +3508,70 @@ func handle_strike_effect(card_id :int, effect, performing_player : Player):
...
-					spaces_to_location -= 1
-
-				if move_min <= spaces_to_location and spaces_to_location <= move_max:
-					decision_info.limitation.append(i)
-					decision_info.choice.append({
-						"effect_type": "move_to_space",
-						"amount": i
-					})
-			events += [create_event(Enums.EventType.EventType_ChooseArenaLocationForEffect, performing_player.my_id, 0)]
+			if len(decision_info.limitation) > 0:
```
daniel-k-taylor, line 3569: maybe account for pass here? I would still want movement blocked to show up if pass/move to own space is the only option.

daniel-k-taylor (dismissed):
I think what you have is probably fine?
If you can just confirm in the UI that the move any behavior looks fine with like arakune's move to any space from corner boost or nu's move to any space legacyedge effect and the 2 movement blocking things.

scenes/game/local_game.gd
daniel-k-taylor: Let's do applying it to other effects as a separate change.

daniel-k-taylor (approved):

PR #57, opened 2024-01-31
Implement some small fixes

Rose-Personal:
nu-13 exceeded character ability changed to set_strike
fix ai stall on naming a card effects
handle defend command losing armor correctly, add tests
cleanup character bonus panel earlier (before overdrives)

propose we leave RandomIterations checked in at 1 since it doesn't run that slowly, nice to have a default sanity check

### DIFF
--- a/scenes/game/enums.gd
+++ b/scenes/game/enums.gd
@@ -136,7 +136,7 @@ enum EventType {
 	EventType_Strike_Miss,
 	EventType_Strike_ChooseToDiscard,
 	EventType_Strike_ChooseToDiscard_Info,
-	EventType_Strike_CleanupUI,
+	EventType_Strike_Cleanup,
 	EventType_Strike_OpponentCantMovePast,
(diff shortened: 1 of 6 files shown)


scenes/game/enums.gd
```diff
@@ -136,6 +136,7 @@ enum EventType {
 	EventType_Strike_Miss,
 	EventType_Strike_ChooseToDiscard,
 	EventType_Strike_ChooseToDiscard_Info,
+	EventType_Strike_CleanupUI,
```
daniel-k-taylor, line 139: This needs to be named Strike_Cleanup or Strike_Completed.

I know it's not super clean throughout the whole thing, but generally LocalGame should not "know" about UI.
I'm particularly bad about this with the strings though.

daniel-k-taylor (changes requested):
Just asking to rename the event.

daniel-k-taylor (approved):

PR #62, opened 2024-02-01
Update local_game.gd

AdwinsM:
added dodge_at_speed and Lumi's wonderful if you're faster you die thing


daniel-k-taylor:
You should close this PR. it seems like you already have these changes in the other PR?

PR #63, opened 2024-02-01
Add UI describing plague knight discards

Rose-Personal:
Maybe this + sidestep could eventually be extensible and add to some kind of array of messages to display in this location? I don't think that would be hard, and could also be good for Critical. Will think about it.

Results of this change:

plague_knight_no_discard

plague_knight_frontside

plague_knight_exceed

### DIFF
--- a/scenes/game/game.gd
+++ b/scenes/game/game.gd
@@ -1826,6 +1826,10 @@ func begin_strike_choosing(strike_response : bool, cancel_allowed : bool,
 	if cards_that_will_not_hit.size() > 0:
 		for card in cards_that_will_not_hit:
 			dialogue += "\n" + card + " will not hit."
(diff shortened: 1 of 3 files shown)


daniel-k-taylor (approved):
A few thoughts:
1) Maybe this would be better on strike_stat_boosts than on player?
2) If we add more of these, I would want it to be some kind of "get_strike_hint_text" or something? I dunno.
3) Critical shows up in the character effect block, but the plague knight things don't?
I'm adding something with kokonoe that lets her gravitron show up in character effect after she pays the set_strike effect.
Overall, this whole UI bit could use some love anyway.
4) How amazing would it be if you could hover over the text of the card and it popped up with the card image?
Apparently there are some thoughts on doing this here:
https://www.reddit.com/r/godot/comments/w34wmd/label_with_certain_words_that_show_a_popup_when/

I think this is fine to put in for now, as we can just change it later, but I don't want to keep adding more and more functions to get individual things for this UI. It would need to be more generalized.

PR #64, opened 2024-02-01
Implementing Chaos

Mach131:
had a bit of fun with the buddy names


scenes/game/local_game.gd
```diff
@@ -477,6 +479,7 @@ class Player:
 	var do_not_cleanup_buddy_this_turn : bool
 	var cannot_move : bool
 	var cannot_move_past_opponent : bool
+	var cannot_move_past_opponent_buddy_id : Variant
```
daniel-k-taylor, line 482: why bother putting Variant?

daniel-k-taylor: i guess since it is assumed string maybe but this is nullable.

scenes/game/local_game.gd
```diff
@@ -1666,12 +1692,29 @@ class Player:
...
 		var previous_location = arena_location
-		var other_location = parent._get_player(parent.get_other_player(my_id)).arena_location
+		var other_player = parent._get_player(parent.get_other_player(my_id))
+		var other_location = other_player.arena_location
 		var new_location
 		if arena_location < other_location:
 			new_location = min(other_location-1, arena_location+amount)
 		else:
 			new_location = max(other_location+1, arena_location-amount)
+
```
daniel-k-taylor, line 1702: I'm starting to worry about the amount of code we're duplicating between close/advance/retreat.
Maybe close/retreat can basically be the same function just in different directions?
And close is really just advance but opponent is blocking you.

I'll let you decide if you want to do this or not.

daniel-k-taylor: I know for Kokonoe I was also adding some stop_on_buddy functionality, which is slightly different from your blocked_by_buddy since it is inclusive.

data/card_definitions.json
```diff
@@ -1974,6 +1974,370 @@
...
+				"timing": "during_strike",
+				"condition": "not_buddy_in_play",
+				"effect_type": "armorup",
+				"amount": 2
+			},
+			{
+				"timing": "hit",
+				"condition": "range",
+				"condition_amount": 0,
+				"condition_from_source": true,
```
daniel-k-taylor, line 2314: maybe instead of this, condition buddy_in_opponent_space used for litchi?

scenes/game/local_game.gd
```diff
@@ -2763,6 +2847,17 @@ func is_effect_condition_met(performing_player : Player, effect, local_condition
 		elif condition == "range":
 			var amount = effect['condition_amount']
 			var distance = abs(performing_player.arena_location - other_player.arena_location)
+
+			var from_source = 'condition_from_source' in effect and effect['condition_from_source']
```
daniel-k-taylor, line 2851: shouldn't need any of this if you just use litchi's thing, buddy_in_opponent_space

data/card_definitions.json
daniel-k-taylor: buddy shouldn't be in play unless they used their char ability

scenes/game/local_game.gd
```diff
@@ -3477,7 +3583,12 @@ func handle_strike_effect(card_id :int, effect, performing_player : Player):
 			events += [create_event(Enums.EventType.EventType_Strike_GuardUp, performing_player.my_id, effect['amount'])]
 		"higher_speed_misses":
 			performing_player.strike_stat_boosts.higher_speed_misses = true
-			_append_log_full(Enums.LogType.LogType_Effect, performing_player, "will dodge attacks of a higher speed!")
+			if 'dodge_at_speed_greater_than' in effect:
+				var speed_dodge = effect['dodge_at_speed_greater_than']
+				performing_player.strike_stat_boosts.dodge_at_speed_greater_than = speed_dodge
+				_append_log_full(Enums.LogType.LogType_Effect, performing_player, "will dodge attacks of speed %s or greater!" % speed_dodge)
```
daniel-k-taylor, line 3589: the effect is greater_than, but this log says speed equal or greater.

scenes/game/local_game.gd
```diff
@@ -5023,13 +5155,20 @@ func in_range(attacking_player, defending_player, card, combat_logging=false):
...
 	# Speed dodge
+	var attacking_speed = get_total_speed(attacking_player, active_strike.get_player_card(attacking_player))
+	var defending_speed = get_total_speed(defending_player, active_strike.get_player_card(defending_player))
 	if defending_player.strike_stat_boosts.higher_speed_misses:
-		var attacking_speed = get_total_speed(attacking_player, active_strike.get_player_card(attacking_player))
-		var defending_speed = get_total_speed(defending_player, active_strike.get_player_card(defending_player))
-		if attacking_speed > defending_speed:
+		var speed_dodge = defending_player.strike_stat_boosts.dodge_at_speed_greater_than
+		if speed_dodge > 0:
+			if attacking_speed >= speed_dodge:
```
daniel-k-taylor, line 5163: effect is named greater than, but this is >=

daniel-k-taylor (changes requested):

scenes/game/game.gd
```diff
@@ -490,7 +492,7 @@ func move_character_to_arena_square(character, arena_location, immediate: bool,
 	var offset_y = $ArenaNode/RowButtons.position.y
 	target_position.y -= character.get_size().y * character.scale.y / 2 + offset_y + character.vertical_offset
 	if buddy_offset != 0:
-		target_position.x += buddy_offset * (character.get_size().x * character.scale.x /4)
+		target_position.x += buddy_offset * (character.get_size().x * character.scale.x /4) + character.horizontal_offset
```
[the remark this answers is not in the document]
Mach131: I didn't mean for the offset to be mirrored, if that's what you mean; the offset's direction seems consistent facing both sides

scenes/game/local_game.gd
Mach131: yeah, I wanted to avoid making two variables for this; maybe that would be clearer?

data/card_definitions.json
Mach131: will borrow the buddy_in_opponent_space condition; however, the buddy can also be in play because of the deep revenance boost

scenes/game/game.gd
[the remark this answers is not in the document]
daniel-k-taylor: and it works for opponent's buddy too?
Just want to double check since there is opportunity for weirdness here

scenes/game/local_game.gd
daniel-k-taylor: it's fine i guess

data/card_definitions.json
daniel-k-taylor: oh weird, so you deep rev strike and if it's this it calculates range from yourself.
But it should still work in that case?

Mach131: yup, it still gets the place in space effects; not sure how useful it is but

scenes/game/game.gd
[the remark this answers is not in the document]
Mach131: yeah, looks the same on both clients

scenes/game/local_game.gd
Mach131: i wonder, is there anything stopping us from having advance/retreat/close just redirect to move_to (with a blocking option for close)?

daniel-k-taylor: Well, you don't know exactly where you're moving to depending on the opponent for advance.
I think the most reasonable is like:
move_in_direction_by_amount(direction : int, amount : int, blocked_by_opponent : bool)
Then it moves you in that direction 1 at a time, checking blocked_by_opponent, blocked_by_buddy_id stuff, and the new stop_on_buddy thing I'm adding.
When you land on the opponent (and you weren't blocked by them), you move a bonus space in direction.

How does that sound?

daniel-k-taylor: Close/Advance/Retreat will then calculate the direction first and pass that in.

Mach131: Probably more simple than what I was trying out; I'll see how well it works

scenes/game/local_game.gd
```diff
@@ -1636,6 +1636,65 @@ class Player:
 			if old_pos == buddy_old_pos:
 				handle_on_buddy_boosts(false)
 
+	func move_in_direction_by_amount(go_left : bool, amount : int, stop_at_opponent : bool, movement_type : String):
+		var events = []
+		var direction = -1 if go_left else 1
```
daniel-k-taylor, line 1641: personally i would have just passed in direction as a 1 or -1 int, but this is less open to passing in weirdo stuff that is not 1 or -1.

scenes/game/local_game.gd
```diff
@@ -1636,6 +1636,65 @@ class Player:
...
+			if cannot_move_past_opponent_buddy_id:
+				var other_buddy_loc = other_player.get_buddy_location(cannot_move_past_opponent_buddy_id)
+				if target_location == other_buddy_loc:
+					movement_shortened = true
+					blocked_by_buddy = true
+					break
+
+			if target_location == other_player_loc:
+				if stop_at_opponent:
+					break
```
daniel-k-taylor, line 1665: Shouldn't target_location then be changed to not be on the opponent?

daniel-k-taylor: oh, it breaks, got it

daniel-k-taylor (approved):
Love it.

PR #65, opened 2024-02-02
Akuma

chisfff:


data/decks/akuma.json
```diff
@@ -0,0 +1,165 @@
...
+				"effect_type": "critical",
+				"character_effect": true
+			},
+			"gauge_max": 1
+		},
+		{
+			"timing": "set_strike",
+			"condition": "is_critical",
+			"effect_type": "powerup_both_players",
+			"amount": 2
```
Mach131, line 22: just checking, does this work properly? wondering if there's a case where the ordering may get mixed up with the crit trigger (if so, changing this to during_strike may be smoother)

scenes/game/local_game.gd
```diff
@@ -3611,6 +3622,10 @@ func handle_strike_effect(card_id :int, effect, performing_player : Player):
 			_append_log_full(Enums.LogType.LogType_CharacterMovement, performing_player, "moves to %s, from space %s to %s." % [buddy_name, str(previous_location), str(performing_player.arena_location)])
 		"multiply_power_bonuses":
 			performing_player.strike_stat_boosts.power_bonus_multiplier = max(effect['amount'], performing_player.strike_stat_boosts.power_bonus_multiplier)
+		"multiply_positive_power_bonuses":
+			var power_bonuses = performing_player.strike_stat_boosts.power
+			if power_bonuses > 0:
```
Mach131, line 3627: I don't think this is quite sufficient; for instance, if you boost fierce but Arakune uses his -1 power boost, strike_stat_boosts.power will be equal to the sum (1) and you'd get a total boost of +2 instead of the +3 you'd want (4 - 1). I was thinking this might be slightly annoying to implement; in some form, you'd want to keep track of positive power bonuses separately from cumulative power (ctrl+f for wherever strike_stat_boosts.power is changed and adjust your new variable if the effect's boost is positive; make sure to also adjust the part of the code that undoes boost effects)

Mach131: Also as a note, this might be the kind of thing that'd be useful to include in unit tests when you get to them

Mach131 (changes requested):
Looks mostly good; a few minor comments, but I think the positive power bonus piece might need a bit more work

data/decks/akuma.json
chisfff: It does work properly, I think it does not matter right now but I can change it if you want.

scenes/game/local_game.gd
chisfff: hmmm with my coding skills i don't think i will be able to accomplish this implementation without a lot of help.

Mach131: It hopefully shouldn't be too hard; mostly just add the variable for tracking positive_power_bonuses, and whenever the normal power boost variable is changed, check if the change is positive and change positive_power_bonuses if so

chisfff: i'll give it a shot

data/card_definitions.json
```diff
@@ -1,4 +1,310 @@
...
+		"effects": [
+			{
+				"timing": "during_strike",
+				"condition": "is_critical",
+				"effect_type": "ignore_armor"
+			},
+			{
+				"timing": "before",
+				"effect_type": "close",
+				"amount": 1
```
daniel-k-taylor, line 192: isn't this supposed to be close 2?

data/card_definitions.json
```diff
@@ -1,4 +1,310 @@
...
+		"speed": 4,
+		"power": 4,
+		"armor": 0,
+		"guard": 0,
+		"effects": [
+			{
+				"timing": "before",
+				"effect_type": "advance",
+				"amount": 2,
+				"bonus_effect": {
```
daniel-k-taylor, line 236: this part is supposed to be a hit effect.
maybe use set_strike_x as the advanced through part and set it to one, and then have a hit effect to discard the set x value?

data/decks/akuma.json
daniel-k-taylor: set_strike should go in order I think?

scenes/game/local_game.gd
```diff
@@ -2765,6 +2765,17 @@ func is_effect_condition_met(performing_player : Player, effect, local_condition
 			return not performing_player.strike_stat_boosts.critical
 		elif condition == "choose_cards_from_top_deck_action":
 			return decision_info.action == effect["condition_details"]
+		elif condition == "total_powerup_greater_or_equal":
+			var amount = effect["condition_amount"]
+			var total_powerup = performing_player.strike_stat_boosts.power
+			return total_powerup >= amount
+		elif condition == "opponent_total_guard_greater_or_equal":
+			var amount = effect["condition_amount"]
+			var card = active_strike.get_player_card(other_player)
```
daniel-k-taylor, line 2774: I don't understand why you can't use the get_total_guard function here.

daniel-k-taylor: if it isn't working, we need to debug and fix it as guard is probably broken somehow

scenes/game/local_game.gd
daniel-k-taylor: Please add a function that is something like add_player_power_bonus(player, amount).
Then make everything that changes the strike stat boosts call that, and record both the current power variable but also add your new variable to track positive bonuses

daniel-k-taylor (changes requested):

scenes/game/local_game.gd
chisfff: you lost me at functions. If anyone of you have time can you implement this for me?

data/card_definitions.json
chisfff: ryu's tatsu is coded the same way but works in the way you think by only gaining advantage when hitting the opponent, when i tested by using hit effect it errored out.

daniel-k-taylor: that's probably wrong too then

Mach131: oh yeah tatsu probably is wrong; i'll look into making this kind of thing easier to write later

chisfff: ok probably a good change since ranged dodges is a thing and don't want ryu to be gaining advantage for free

daniel-k-taylor:
Closed this as I accidentally made a new branch

PR #66, opened 2024-02-02
Kokonoe

daniel-k-taylor:


scenes/game/local_game.gd
```diff
@@ -5560,6 +5837,154 @@ func strike_send_attack_to_discard_or_gauge(performing_player : Player, card):
...
+	# The player is mid-strike and playing an extra attack card.
+	# These are the steps of extra attack:
+	# * Pay its cost.
+	# * Put it into play, gaining any Armor, Guard, and passive effects on it.
+	# * Perform its Before effects, then check its Range.
+	# * If it hit, perform its Hit effects.
+	# * If it hit, deal damage equal to its printed Power.
+	# * Perform its After effects.
+	# * At the end of the Strike, if the additional attack hit, add it to your Gauge. Otherwise, discard it.
+	# * Note that bonuses and effects that apply to your attack (e.g., Continuous Boost effects such as "+2 Power") only apply to your initial attack unless they specify "Your attacks have [...]" (as the Graviton does).
```
Mach131, line 5851: hopefully this isn't something we have to find and support for other boosts retroactively?

scenes/game/local_game.gd
```diff
@@ -5560,6 +5837,154 @@ func strike_send_attack_to_discard_or_gauge(performing_player : Player, card):
...
+				if game_over:
+					change_game_state(Enums.GameState.GameState_GameOver)
+				else:
+					# Intentional events = because events are passed in.
+					events = continue_resolve_strike(events)
+				break
+	return events
+
+func determine_if_attack_hits(events, attacker_player : Player, defender_player : Player, card : GameCard):
+	# events actually is used, but there's an unused parameter warning for some reason, so add _.
```
Mach131, line 5962: is this an outdated comment?

test/unit/test_randomai.gd
```diff
@@ -1,7 +1,7 @@
 extends GutTest
 
 # Leave at 1 checked in so someone doesn't accidentally run all tests at 100.
-const RandomIterations = 1
+const RandomIterations = 0
```
Mach131, line 4: reminder to set to 1

test/unit/test_randomai.gd
```diff
@@ -1,7 +1,7 @@
 extends GutTest
```
Mach131, line 1: This currently just sets up a bunch of mirror matches, right? I wonder if it would be helpful to add a special test that runs two different random characters against each other; it might help to find a few more strange corner cases, if annotated properly

Mach131 (approved):
Seems reasonable to me if the tests pass, just a bunch of text/formatting-related comments mostly; very hard to envision the exact ways in which things might fall apart, hence the randomai tests suggestion

data/card_definitions.json
```diff
@@ -6914,6 +6914,416 @@
...
+		"speed": 4,
+		"power": 4,
+		"armor": 0,
+		"guard": 0,
+		"effects": [
+			{
+				"timing": "before",
+				"effect_type": "advance",
+				"stop_on_buddy_space": "gravitron",
+				"description": "Advance 4. If you enter Gravitron's space, stop and gain advantage.",
```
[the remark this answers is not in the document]
daniel-k-taylor: no, so i added that

data/decks/kokonoe.json
```diff
@@ -0,0 +1,222 @@
...
+	"on_exceed": {
+		"effect_type": "take_bonus_actions",
+		"amount": 1,
+		"and": {
+			"effect_type": "place_buddy_in_any_space",
+			"buddy_name": "Gravitron"
+		}
+	},
+	"overdrive_effect": {
+		"effect_type": "place_buddy_in_any_space",
```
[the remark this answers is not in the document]
daniel-k-taylor: the overdrive one isn't, so i'll make it a choice

scenes/game/local_game.gd
```diff
@@ -1749,21 +1813,21 @@ class Player:
...
-			events += move_in_direction_by_amount(false, amount, true, "close")
+			events += move_in_direction_by_amount(false, amount, true, -1, "close")
 		else:
-			events += move_in_direction_by_amount(true, amount, true, "close")
+			events += move_in_direction_by_amount(true, amount, true, -1, "close")
 
 		return events
 
-	func advance(amount):
+	func advance(amount, stop_on_space):
```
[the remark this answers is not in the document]
daniel-k-taylor: There is only 1 place that calls this, so it might as well not be optional

scenes/game/local_game.gd
daniel-k-taylor: I think we just have to do a survey of all the things a given character that has extra attacks can boost.

scenes/game/local_game.gd
daniel-k-taylor: yes, this was me being confused about a warning, which was actually a real bug.
So, if you use events += [] inside this function, this does a new assignment and loses the reference to events and doesn't actually add to the array.
So if we ever aren't returning events from a function we pass events into, it needs to use append_array/append.
Potentially we should go change this everywhere anyway.

test/unit/test_randomai.gd
daniel-k-taylor: Sure, we could have a test that just runs every character against every other character.
Probably take a while to run though, but it would have at least caught stuff like hazama vs critical

Mach131 (approved):

PR #68, opened 2024-02-03
Add public hand tracking

daniel-k-taylor:


Mach131 (approved):
Makes sense to me; we'll probably hear back from people if there's anything that feels unintuitive about it

PR #69, opened 2024-02-03
Akuma

daniel-k-taylor:


scenes/game/local_game.gd
```diff
@@ -2031,10 +2041,10 @@ class Player:
 							strike_stat_boosts.dodge_at_range_max = -1
 							strike_stat_boosts.dodge_at_range_from_buddy = false
 					"powerup":
-						strike_stat_boosts.power -= effect['amount']
+						add_power_bonus(-effect['amount'])
 					"powerup_both_players":
-						strike_stat_boosts.power -= effect['amount']
-						opposing_player.strike_stat_boosts.power -= effect['amount']
+						add_power_bonus(-effect['amount'])
```
Mach131, line 2046: won't this not properly update the positive power bonuses (because it's being passed a negative value for positive bonuses)? might be easier to have a separate remove_power_bonus function, or a second parameter

scenes/game/local_game.gd
```diff
@@ -5870,6 +5907,7 @@ func begin_extra_attack(events, performing_player : Player, card_id : int):
 
 	# Other relevant passives that need to be undone.
 	# If other extra attack characters are added, consider what else might be needed.
+	# For example: If an extra attack character can have positive power bonus boosts, those need to be removed too.
```
Mach131, line 5910: thankfully I think dan is the only other one in 3-7 and he doesn't seem to care about positive power

Mach131 (changes requested):
Seems good, just one important point about reversing power bonuses

scenes/game/local_game.gd
daniel-k-taylor: oh, yea

Mach131 (approved):

PR #78, opened 2024-02-06
Implementing Propeller Knight

Mach131:
mostly just copying rose's changes


scenes/game/local_game.gd
```diff
@@ -1842,8 +1851,8 @@ class Player:
 			parent._append_log_full(Enums.LogType.LogType_CardInfo, self, "reveals the top card of their deck to the opponent.")
 		else:
 			parent._append_log_full(Enums.LogType.LogType_CardInfo, self, "reveals the top card of their deck: %s." % card_name)
+			public_topdeck_id = deck[0].id
```
daniel-k-taylor, line 1854: I think this should still happen in both cases?
The public state should be the same on both clients.

scenes/game/local_game.gd
```diff
@@ -3234,6 +3245,18 @@ func is_effect_condition_met(performing_player : Player, effect, local_condition
 			return true
 		elif condition == "opponent_at_edge_of_arena":
 			return other_player.is_at_edge_of_arena()
+		elif condition == "opponent_at_max_range":
+			assert(active_strike)
+			var attack_card = active_strike.get_player_card(performing_player)
+			var max_range = get_card_stat(performing_player, attack_card, 'range_max') + performing_player.strike_stat_boosts.max_range
```
daniel-k-taylor, line 3251: probably a good candidate for a get_total_max_range function. And then technically we should probably overwrite range in attack removal cleanup.
But does this matter for any characters? Not sure if anyone cares about the range of their opponent's attack?
Maybe don't bother if nobody cares.

scenes/game/local_game.gd
```diff
@@ -4975,6 +4998,8 @@ func handle_strike_effect(card_id :int, effect, performing_player : Player):
 			performing_player.life = amount
 			events += [create_event(Enums.EventType.EventType_Strike_GainLife, performing_player.my_id, amount, "", performing_player.life)]
 			_append_log_full(Enums.LogType.LogType_Health, performing_player, "gains %s life, bringing them to %s!" % [str(amount), str(performing_player.life)])
+		"shuffle_discard_in_place":
+			performing_player.random_shuffle_discard_in_place()
```
daniel-k-taylor, line 5002: hmm, do we need an event for this?
Otherwise the top discard card won't be correct i think?
Probably need some event to reorder the discards.

daniel-k-taylor (changes requested):

scenes/game/local_game.gd
Mach131: The issue I ran into was that Anji could peek at your topdeck, and if you have a copy of the same card in your hand you can see the icon there

scenes/game/local_game.gd
Mach131: Maybe carl? but only if the attack could somehow be removed before it hit? but yeah might at least make get_total_max_range, and probably get_attack_origin while i'm at it

scenes/game/local_game.gd
Mach131: sorry, what do you mean by top discard card being incorrect?

scenes/game/local_game.gd
```diff
@@ -5820,15 +5801,14 @@ func do_effects_for_timing(timing_name : String, performing_player : Player, car
 func is_location_in_range(attacking_player, card, test_location : int):
 	if get_card_stat(attacking_player, card, 'range_min') == -1 or attacking_player.strike_stat_boosts.overwrite_range_to_invalid:
```
daniel-k-taylor, line 5808: I feel like maybe the get_total_min/max should account for this?

daniel-k-taylor: Probably not needed for this change though.

daniel-k-taylor (approved):

scenes/game/local_game.gd
Mach131: Off the top of my head I think it's always just either only you or only the opponent? unsure if anything similar happens for "look at the top card of your deck" type effects

PR #81, opened 2024-02-08
Mole knight

Mach131:


data/card_definitions.json
```diff
@@ -10457,6 +10457,324 @@
...
+		"boost": {
+			"boost_type": "immediate",
+			"force_cost": 0,
+			"cancel_cost": -1,
+			"display_name": "Lost Treasure",
+			"effects": [
+				{
+					"timing": "immediate",
+					"effect_type": "choose_discard",
+					"limitation": "special/ultra",
```
daniel-k-taylor, line 10585: I probably would have done special_or_ultra, but this is fine i guess.

data/card_definitions.json
```diff
@@ -10457,6 +10457,324 @@
...
+			"display_name": "Come At Me!",
+			"effects": [
+				{
+					"timing": "during_strike",
+					"effect_type": "powerup",
+					"amount": 3
+				},
+				{
+					"timing": "start_of_next_turn",
+					"effect_type": "add_to_gauge_immediately"
```
daniel-k-taylor, line 10646: This should have "and draw a card"  too

data/decks/mole.json
```diff
@@ -0,0 +1,238 @@
...
+		}
+	}],
+	"ability_effects": [
+		{
+			"timing": "set_strike",
+			"condition": "initiated_strike",
+			"effect_type": "choice",
+			"choice": [
+				{
+					"effect_type": "set_used_character_bonus",
```
daniel-k-taylor, line 46: Wouldn't add_attack_effect work? Like Kokonoe?

data/decks/mole.json
```diff
@@ -0,0 +1,238 @@
...
+					"character_effect": true,
+					"linked_effect": {
+						"timing": "before",
+						"condition": "buddy_space_unoccupied",
+						"condition_detail": "Burrow",
+						"effect_type": "move_to_buddy",
```
daniel-k-taylor, line 85: how does this interact with cannot move effects?
The text says "If you did, +2 power"
Did what? Tried to advance or actually advanced into the burrow?

data/card_definitions.json
daniel-k-taylor: in fact, yours probably works better if you want to just use the string as is.

daniel-k-taylor (changes requested):

data/card_definitions.json
Mach131: yeah, mostly went with this for the string appearance

data/card_definitions.json
Mach131: oh huh, this version is better than the S6 version

data/decks/mole.json
Mach131: Should be the same thing; i guess it avoids the need to repeat the effect

data/decks/mole.json
Mach131: I checked, you only get the power if you successfully advance/retreat into the burrow (hence the on_buddy_space check)

daniel-k-taylor (approved):

PR #83, opened 2024-02-08
Seth

Mach131:


scenes/game/local_game.gd
```diff
@@ -8112,8 +8138,11 @@ func do_bonus_turn_action(performing_player : Player, action_index : int):
 	# Do the bonus action effects.
 	var events = []
 	events += handle_strike_effect(chosen_action['card_id'], chosen_action, performing_player)
-	if game_state != Enums.GameState.GameState_WaitForStrike:
+	if game_state not in [Enums.GameState.GameState_WaitForStrike, Enums.GameState.GameState_PlayerDecision]:
```
daniel-k-taylor, line 8141: I feel like maybe we should be calling the continue player resolution function I added recently?

daniel-k-taylor: If we have a more complicated bonus action later with multiple decisions, we can have this put the extra effects into the remaining_character_action_effects and it probably "just works"?

daniel-k-taylor: In fact, maybe this function should just put the effects in there, set active_character_action to true, and call the continue resolution function?
that would then handle calling do_remaining... and calling to handle the strike effects, etc.

daniel-k-taylor (changes requested):

scenes/game/local_game.gd
Mach131: this seems to work well!

daniel-k-taylor (approved):

PR #84, opened 2024-02-09
Add observer mode

daniel-k-taylor:


scenes/menu/main_menu.gd
```diff
@@ -315,3 +374,25 @@ func _on_bgm_check_box_toggled(button_pressed : bool):
...
+	room_select.text = selected_match['name']
+	_on_join_button_pressed()
+
+func _on_modal_list_observe_match_pressed(row_index):
+	var matches = NetworkManager.get_match_list()
+	var selected_match = matches[row_index]
+	var room_name = selected_match['name']
+	var player_name = get_player_name()
+	NetworkManager.observe_room(player_name, room_name)
+	update_buttons(true)
```
Mach131, line 397: Would it be too much extra effort to support people opting out of being observed? (e.g. players can press a checkbox before entering a room/matchmaking; if either player has it enabled then the match object stores a flag that this checks)

Mach131 (commented):
Seems good at a high level; I just remember it being brought up that this being optional would likely be a good idea, if that's feasible

scenes/menu/main_menu.gd
daniel-k-taylor: do we really need that? I don't think I would allow it for the matchmaking queue. And nobody really even uses rooms.

Mach131: I do see a fair number of rooms occasionally, but yee fair enough

Mach131 (approved):

PR #86, opened 2024-02-10
Ordering effects with unmet conditions

Mach131:
Also includes some smaller changes (ex power counting towards doubling, a few boosts causing strikes without going through the normal boost logic, c. viper's portrait missing the background)


test/unit/test_arakune.gd
```diff
@@ -315,6 +315,7 @@ func test_arakune_exceed_and_strike_with_bonus_wildswing_finverse_wildswing():
 	# Pay cost wild swing into assault
 	assert_true(game_logic.do_pay_strike_cost(player1, [], true))
 	# p2 hits with grasp
+	handle_simultaneous_effects(player1, player2, [])
```
daniel-k-taylor, line 318: Oh, the other effect is his unmet character ability? I wonder if we should somehow split this into conditions that could possibly change, or conditions that can't possibly change during the battle.
Like, having to resolve simultaneous conditions for Critical and stuff is going to be dumb. Your attack will never become critical, or switch from normal to speciall.

daniel-k-taylor: is_critical/is_not_critical
was_hit
initiated_strike/not_initiated_strike
exceeded/not_exceeded (I don't think anybody can exceed during a phase that can have an effect on that phase)
buddy_in_play - probably safe to skip
used_character_bonus
used_character_action
hit_opponent
opponent_stunned - Only rachel maybe does damage in after, and doesn't have this condition?
initiated_face_up
stunned/not_stunned
initiated_after_moving
was_wild_swing
last_turn_was_strike
speed_greater_than
is_special_or_ultra_attack/is_normal_attack/is_special_attack
discarded_matches_attack_speed
canceled_this_turn

I did a survey and I think that's all stuff that can't possibly change to being met somehow during strike

daniel-k-taylor (changes requested):
I think this is good, I just think we should try to avoid prompting simul choices if we know some conditions can never be met during the strike.

scenes/game/game.gd
```diff
@@ -2323,7 +2323,7 @@ func _on_effect_choice(event):
 				if effect_met:
 					extra_choice_text.append("")
 				else:
-					extra_choice_text.append("[color=red][lb]Condition Unmet[rb][/color] ")
+					extra_choice_text.append("[color=red][lb]FAILS[rb][/color] ")
```
daniel-k-taylor, line 2326: Maybe just FAIL? no S

daniel-k-taylor (changes requested):

daniel-k-taylor (approved):

PR #87, opened 2024-02-10
Zangief

daniel-k-taylor:


data/card_definitions.json
```diff
@@ -19359,6 +19359,317 @@
...
+		"power": 8,
+		"armor": 0,
+		"guard": 6,
+		"effects": [
+			{
+				"timing": "during_strike",
+				"effect_type": "opponent_cant_move_past"
+			},
+			{
+				"timing": "now",
```
Mach131, line 19424: this probably doesn't need the now timing effect since it's on an attack rather than a boost (does the discarded timing work for attacks? should it be cleanup instead?)

data/card_definitions.json
```diff
@@ -19359,6 +19359,317 @@
...
+			{
+				"timing": "during_strike",
+				"effect_type": "ignore_push_and_pull"
+			},
+			{
+				"timing": "after",
+				"condition": "is_critical",
+				"effect_type": "move_to_any_space",
+				"move_min": 1,
+				"move_max": 2
```
Mach131, line 19616: we should probably go through and update more of these eventually

scenes/game/local_game.gd
```diff
@@ -4318,6 +4333,94 @@ func handle_strike_effect(card_id :int, effect, performing_player : Player):
...
+			decision_info.clear()
+			decision_info.type = Enums.DecisionType.DecisionType_PickNumberFromRange
+			decision_info.player = performing_player.my_id
+			decision_info.choice_card_id = card_id
+			decision_info.choice = []
+			decision_info.limitation = []
+			if effect['target_effect'] == "opponent_discard_range_or_reveal":
+				decision_info.amount_min = 0
+				decision_info.amount = 9
+				decision_info.valid_zones = ["Range X", "Range -"]
```
Mach131, line 4346: would "Range N/A" be clearer than "Range -"?

scenes/game/local_game.gd
```diff
@@ -4318,6 +4333,94 @@ func handle_strike_effect(card_id :int, effect, performing_player : Player):
...
+							card_ids_in_range.append(card.id)
+				else:
+					assert(false, "Unknown target range")
+			else:
+				# If the range is an actual number.
+				for card in opposing_player.hand:
+					var card_range_min = card.definition['range_min']
+					var card_range_max = card.definition['range_max']
+					# Some cards have mixed numbers and strings.
+					# Potentially evaluate? Currently only Phonon though and max==min if not in strike.
```
Mach131, line 4392: If you name, for example, 3 when you're at range 3 against Nine, would Kunzite be considered in range? or does that only get targeted if you name Range X? if the former, it might be helpful to attempt to resolve the ranges (maybe the range resolution function should have logic similar to this if a strike isn't active, in either case)

scenes/game/local_game.gd
```diff
@@ -4318,6 +4333,94 @@ func handle_strike_effect(card_id :int, effect, performing_player : Player):
...
+					decision_info.extra_info = "include Range %s" % target_range
+				decision_info.choice = card_ids_in_range
+				decision_info.can_pass = false
+
+				decision_info.choice_card_id = card_id
+				decision_info.player = opposing_player.my_id
+				events += [create_event(Enums.EventType.EventType_Strike_ChooseToDiscard, opposing_player.my_id, amount)]
+			else:
+				# Didn't have any that matched, so forced to reveal hand.
+				events += opposing_player.reveal_hand()
```
Mach131, line 4423: Maybe a small log message here would be nice (and also one mentioning the range that the player named, for both cases)

Mach131 (changes requested):
Zangief looks good apart from one thing on siberian blizzard; also still a bit unsure about how range parries work

data/card_definitions.json
daniel-k-taylor: hmm, I think putting remove on cleanup is probably what we want also?

scenes/game/local_game.gd
```diff
@@ -4390,6 +4394,10 @@ func handle_strike_effect(card_id :int, effect, performing_player : Player):
 					var card_range_max = card.definition['range_max']
 					# Some cards have mixed numbers and strings.
 					# Potentially evaluate? Currently only Phonon though and max==min if not in strike.
+					if card_range_min is String and card_range_min == "RANGE_TO_OPPONENT":
+						card_range_min = performing_player.distance_to_opponent()
+					if card_range_max is String and card_range_max == "RANGE_TO_OPPONENT":
+						card_range_max = performing_player.distance_to_opponent()
```
Mach131, line 4400: Would it not be better to use get_card_stats? (not sure if there's anything else it matters for right now, but thinking ahead to Enchantress Homing Orb, specifically)

Mach131 (approved):

Mach131 (approved):

PR #92, opened 2024-02-13
Londrekia

Mach131:
Unit tests when I wake up


globals/card_definitions.gd
```diff
@@ -923,6 +964,8 @@ func get_effect_type_text(effect, card_name_source : String = "", char_effect_pa
 				effect_str += "+" + str(effect['amount']) + " Speed per EVERY boost in play."
 			else:
 				effect_str += "+" + str(effect['amount']) + " Speed per boost in play."
+		"spend_all_gauge_and_save_amount":
+			effect_str += "Discard all (X) cards in gauge"
```
daniel-k-taylor, line 968: why X?
Isn't it just all?

scenes/game/local_game.gd
```diff
@@ -3854,8 +4014,13 @@ func handle_strike_effect(card_id :int, effect, performing_player : Player):
 			# This effect is expected to be mid-strike.
 			assert(active_strike)
 			var card_name = card_db.get_card_name(card_id)
-			_append_log_full(Enums.LogType.LogType_Effect, performing_player, "boosts and sustains %s." % card_name)
 			performing_player.strike_stat_boosts.move_strike_to_boosts = true
+			if 'sustain' in effect:
```
daniel-k-taylor, line 4018: there were a bunch of effects using boost_this_then_sustain already that expect sustain to be the default.
Does this change that behavior? Did you go back and change those? I only scrolled through lond's changes in card defs.
I'd prefer to maybe just update all the cards and perhaps even change the name of this effect to something less leading?

daniel-k-taylor: Oh, i see move_strike_to_boosts_sustain defaults to true.
Not sure how I feel about that...
I think at least put a comment here because otherwise we'll just never know what's going on when we come back later.

daniel-k-taylor: maybe even make the extra thing put in the card definition "block_sustain": true? or "dont_sustain" or "dont_actually_sustain" haha

scenes/game/local_game.gd
```diff
@@ -5014,11 +5364,14 @@ func handle_strike_effect(card_id :int, effect, performing_player : Player):
 			for i in range(MinArenaLocation, MaxArenaLocation + 1):
 				if is_location_in_range(performing_player, attack_card, i):
 					decision_info.limitation.append(i)
-					decision_info.choice.append({
+					var new_choice = {
```
daniel-k-taylor, line 5367: was this leftover changes from your previous version?
I guess it's fine to leave in (and above in "place_buddy_into_space", but nothing is using this now right?
maybe also just turn it into an "and"?
Then above in "place_buddy_into_space" you don't actually have to do anything since the and will just process normally. You can remove the if 'additional_effect' code

scenes/game/local_game.gd
```diff
@@ -5233,6 +5586,11 @@ func handle_strike_effect(card_id :int, effect, performing_player : Player):
 			events += handle_strike_effect(card_id, discard_effect, performing_player)
 		"power_modify_per_buddy_between":
 			performing_player.strike_stat_boosts.power_modify_per_buddy_between += effect['amount']
+		"powerdown_per_armor_used":
+			var power_change = - performing_player.strike_stat_boosts.consumed_armor
+			if power_change < 0:
```
daniel-k-taylor, line 5591: i commented when i saw this effect named, but it does look like you could have "amount" in the effect and keep it in the naming pattern of powerup_*

Just makes it a little more reusable later even if we don't actually need it.

daniel-k-taylor (changes requested):
Mostly minor feedback.
I'm happy with having you check this in after the usual CR updates. I just won't spin a build with unfinished rachel.
If we want a build sooner than whenever I get around to it, we could disable her in the UI/random.

scenes/game/local_game.gd
```diff
@@ -5181,14 +5162,13 @@ func handle_strike_effect(card_id : int, effect, performing_player : Player):
...
 					events += opposing_player.discard(discard_ids)
-		"opponent_in_boost_space":
-			var this_boost_location = performing_player.get_boost_location(card_id)
-			return opposing_player.is_in_location(this_boost_location)
 		"pass":
 			# Do nothing.
 			pass
 		"place_boost_in_space":
-			# Currently assumes any space is valid.
+			# TODO
```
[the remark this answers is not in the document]
Mach131: Oh forgot to remove this

globals/card_definitions.gd
Mach131: kind of an artifact from an older implementation whoops

globals/card_definitions.gd
```diff
@@ -20,6 +20,8 @@ func get_random_deck(season : int):
 	else:
 		var season_decks = []
```
daniel-k-taylor, line 21: need to do it for the season == -1 random too.

daniel-k-taylor (changes requested):

daniel-k-taylor (approved):

PR #95, opened 2024-02-15
Rachel finish

daniel-k-taylor:


scenes/game/local_game.gd
```diff
@@ -7844,6 +7869,7 @@ func begin_extra_attack(events, performing_player : Player, card_id : int):
 	active_strike.extra_attack_data.extra_attack_card = card_db.get_card(card_id)
 	active_strike.extra_attack_data.extra_attack_player = performing_player
 	active_strike.extra_attack_data.extra_attack_previous_attack_power_bonus = performing_player.strike_stat_boosts.power
+	active_strike.extra_attack_data.extra_attack_previous_attack_speed_bonus = performing_player.strike_stat_boosts.speed
```
Mach131, line 7872: Should we just go ahead and include armor/guard? I believe the only situation where it matters in any way is like having defend up and hitting that one mole knight lava armor boost, but

Mach131 (approved):
I probably should have reviewed the localgame changes before rebasing on top of them, but the parts that Londrekia used worked fine and the changes here seem good; hopefully the unit tests and Rose's comments should cover everything else

scenes/game/local_game.gd
daniel-k-taylor: armor and guard actually get added to your existing armor/guard according to the faq about extra attacks.
I can't remember now if i checked that I did this correctly though.

PR #96, opened 2024-02-17
Various UI adjustments

Mach131:
- Updating character names
- Making more character effects show up on strike set (and changing some to use add_attack_effect)
- Having card references account for secret sealed areas
- Adjusting font size and centering character names in character select
- Reducing font size of bonus effect panel and supporting wrapping, along with smaller improvements
- Highlighting the names of top-deck cards in relevant choices
- Attempting to alleviate the issues with popouts behaving erratically when more than one thing tries to open/close them (hard to test on my computer though)

### DIFF
--- a/globals/card_definitions.gd
+++ b/globals/card_definitions.gd
@@ -193,6 +193,8 @@ func get_timing_text(timing):
 			text += "When you initiate a strike, "
 		"on_reveal":
 			text += ""
+		"on_seal":
+			text += ""
 		"start_of_next_turn":
 			text += "At start of next turn: "
 		"set_strike":
@@ -902,6 +904,8 @@ func get_effect_type_text(effect, card_name_source : String = "", char_effect_pa
 				effect_str += "Return this to hand"
 		"return_all_cards_gauge_to_hand":
 			effect_str += "Return all cards in gauge to hand."
+		"return_sealed_with_same_speed":
+			effect_str += "Return a sealed card with the same speed to hand."
 		"reveal_copy_for_advantage":
 			effect_str += "Reveal a copy of this attack to Gain Advantage"
 		"reveal_hand":
--- a/globals/card_definitions.gd
+++ b/globals/card_definitions.gd
@@ -6,6 +6,8 @@ var card_definitions_path = "res://data/card_definitions.json"
 var decks_path = "res://data/decks"
 var decks = []
 
+const CardHighlightColor = "#7DF9FF"
+
 func get_deck_test_deck():
 	for deck in decks:
 		if deck['id'] == "rachel":
@@ -427,7 +429,7 @@ func get_effect_type_text(effect, card_name_source : String = "", char_effect_pa
 		"add_top_deck_to_gauge":
 			var topdeck_card = ""
 			if 'card_name' in effect:
-				topdeck_card = "(%s) " % effect['card_name']
+				topdeck_card = "([color=%s]%s[/color]) " % [CardHighlightColor, effect['card_name']]
 			effect_str += "Add top card of deck %sto gauge" % topdeck_card
 		"add_top_discard_to_gauge":
 			effect_str += "Add top card of discard pile to gauge"
@@ -564,7 +566,7 @@ func get_effect_type_text(effect, card_name_source : String = "", char_effect_pa
 			effect_str += "Discard a card from the top of the opponent's deck"
 		"discard_topdeck":
 			if 'card_name' in effect:
-				effect_str += "Discard %s from the top of your deck" % effect['card_name']
+				effect_str += "Discard [color=%s]%s[/color] from the top of your deck" % [CardHighlightColor, effect['card_name']]
 			else:
 				effect_str += "Discard a card from the top of your deck"
 		"discard_random":
@@ -965,7 +967,7 @@ func get_effect_type_text(effect, card_name_source : String = "", char_effect_pa
 				effect_str += "Seal this"
 		"seal_topdeck":
 			if 'card_name' in effect:
-				effect_str += "Seal %s from the top of your deck" % effect['card_name']
+				effect_str += "Seal [color=%s]%s[/color] from the top of your deck" % [CardHighlightColor, effect['card_name']]
 			else:
 				effect_str += "Seal the top card of your deck"
 		"self_discard_choose":
@@ -1016,7 +1018,8 @@ func get_effect_type_text(effect, card_name_source : String = "", char_effect_pa
 		"strike_wild":
 			effect_str += "Wild swing"
 			if 'card_name' in effect:
-				effect_str += " (%s on top of deck)" % effect['card_name']
+				# TODO: repeat for similar effects, test
+				effect_str += " ([color=%s]%s[/color] on top of deck)" % [CardHighlightColor, effect['card_name']]
 		"strike_faceup":
 			effect_str += "Strike face-up"
 		"strike_opponent_sets_first":
--- a/globals/card_definitions.gd
+++ b/globals/card_definitions.gd
@@ -6,7 +6,7 @@ var card_definitions_path = "res://data/card_definitions.json"
 var decks_path = "res://data/decks"
 var decks = []
 
-const CardHighlightColor = "#7DF9FF"
+const CardHighlightColor = "#7DF9FF" # Light blue
 
 func get_deck_test_deck():
 	for deck in decks:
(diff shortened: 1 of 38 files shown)


scenes/game/game.gd
```diff
@@ -1179,6 +1179,8 @@ func _on_advance_turn():
 
 	clear_selected_cards()
 	close_popout()
+	player_bonus_panel.visible = false
```
daniel-k-taylor, line 1184: This feels like something that should happen on the EventType_Strike_Cleanup event handler.

scenes/game/game.gd
```diff
@@ -2908,6 +2910,8 @@ func _update_buttons():
 
 	var action_buttons_visible = ui_state == UIState.UIState_PickTurnAction
 	if action_buttons_visible:
+		player_bonus_panel.visible = false
```
daniel-k-taylor, line 2929: Is this necessary? When is this needed that wasn't taken care of by strike cleanup/advance turn?

globals/card_definitions.gd
```diff
@@ -401,7 +401,7 @@ func get_effect_type_text(effect, card_name_source : String = "", char_effect_pa
 				if char_effect_panel:
 					effect_str += get_effect_text(effect['added_effect'], false, false, false, card_name_source, false)
 				else:
-					effect_str += "Add effect:\n" + get_effect_text(effect['added_effect'], false, false, false, card_name_source, false)
+					effect_str += "Add effect -- " + get_effect_text(effect['added_effect'], false, false, false, card_name_source, false)
```
daniel-k-taylor, line 406: why double dash and not just : or -?
I assume you removed the \n because whatever this was printed in (the choice box?) looks better?

scenes/game/game.gd
```diff
@@ -2418,21 +2419,31 @@ func _on_strike_card_activation(event):
...
+	var effect = event['extra_info']
+	var label_text : String = ""
+	label_text += CardDefinitions.get_effect_text(effect, false, true, true, "", true)
+	_add_bonus_label_text(player, label_text)
+
+func _add_bonus_label_text(player, new_text : String):
 	var bonus_panel = player_bonus_panel
 	var bonus_label = player_bonus_label
-	if not is_player:
+	if player != Enums.PlayerId.PlayerId_Player:
```
daniel-k-taylor, line 2431: player == Enums.PlayerId.PlayerId_Opponent?

scenes/game/game.gd
```diff
@@ -2868,6 +2879,7 @@ func _handle_events(events):
 			Enums.EventType.EventType_Strike_Response:
 				_on_strike_started(event, false)
 			Enums.EventType.EventType_Strike_Response_Ex:
+				_add_bonus_label_text(event['event_player'], "EX Strike")
```
daniel-k-taylor, line 2883: Rather than doing this, I think I'd prefer adding like some wordart "EX" on the top right corner of the card that pops up.
Though to do that we should also probably have that show up for boosts that make it EX and then it also needs to go away if it becomes not EX.
This is probably fine for now? But if you want to make it look really good, probably do this?

scenes/game/game.gd
```diff
@@ -2891,6 +2904,8 @@ func _handle_events(events):
 				delay = _on_damage(event)
 			Enums.EventType.EventType_Strike_WildStrike:
 				_on_strike_started(event, false)
```
daniel-k-taylor, line 2907: what if you did an ex strike that you can't pay then becomes wild? You would need to get rid of your current EX Strike thing.
Another reason to maybe have it displayed on the Card object somehow.

scenes/game/game.gd
```diff
@@ -39,6 +39,8 @@ const LocationInfoButtonPair = preload("res://scenes/game/location_infobutton_pa
 @onready var emote_dialog : EmoteDialog = $EmoteDialog
 @onready var modal_dialog : ModalDialog = $ModalDialog
 
+@onready var PopoutLock = Mutex.new()
```
daniel-k-taylor, line 42: not a huge fan of throwing locks onto things without really understanding the problem.
What threads do you think this is an issue with?
The UI should all run on the UI thread anyway, so I don't see how this is going to solve anything.
Does _process handling events maybe simultaneously do something while the UI thread is processing actually showing the thing or something?

The card popout is definitely one of the worst/hackiest things i wrote in this project.

I think this lock probably? won't break anything.
Have you tried running the game in browser mode from godot? There's a button up there to launch it that way

globals/card_definitions.gd
```diff
@@ -1016,7 +1018,8 @@ func get_effect_type_text(effect, card_name_source : String = "", char_effect_pa
 		"strike_wild":
 			effect_str += "Wild swing"
 			if 'card_name' in effect:
-				effect_str += " (%s on top of deck)" % effect['card_name']
+				# TODO: repeat for similar effects, test
```
daniel-k-taylor, line 1021: what is this todo?

globals/card_definitions.gd
```diff
@@ -6,6 +6,8 @@ var card_definitions_path = "res://data/card_definitions.json"
 var decks_path = "res://data/decks"
 var decks = []
 
+const CardHighlightColor = "#7DF9FF"
```
daniel-k-taylor, line 9: what color is this? comment?

scenes/game/game.gd
```diff
@@ -3130,6 +3151,11 @@ func _update_buttons():
 	action_menu.set_choices(current_instruction_text, button_choices, ultra_force_toggle, instructions_number_picker_min, instructions_number_picker_max, ex_discard_order_toggle)
 	current_action_menu_choices = button_choices
 
+func _choice_text_without_tags(choice_text):
+	var regex = RegEx.new()
```
daniel-k-taylor, line 3155: can you save this regex somewhere that it doesn't create a new one every time it builds the choice strings?

globals/card_definitions.gd
```diff
@@ -427,7 +429,7 @@ func get_effect_type_text(effect, card_name_source : String = "", char_effect_pa
 		"add_top_deck_to_gauge":
 			var topdeck_card = ""
 			if 'card_name' in effect:
-				topdeck_card = "(%s) " % effect['card_name']
+				topdeck_card = "([color=%s]%s[/color]) " % [CardHighlightColor, effect['card_name']]
```
daniel-k-taylor, line 432: I really really want to be able to have this hover show the card, but I'm sure that's going to be a lot of work, haha.

daniel-k-taylor (changes requested):

scenes/game/game.gd
Mach131: I added a notice for Nine's seal effect, but she has one boost that lets her take a bonus action after sealing; I think that's the only case though, so I could just not do that if you think it would be cleaner

scenes/game/game.gd
Mach131: Oh I think this was also added because of a Nine boost (see below)

globals/card_definitions.gd
Mach131: Yeah changed for the choice box, and things like "Add Effect: Hit: Do whatever" looked awkward; I can try single dash

globals/card_definitions.gd
Mach131: whoops

globals/card_definitions.gd
Mach131: light blue

globals/card_definitions.gd
Mach131: the color being hyperlink-blue probably doesn't help, admittedly

scenes/game/game.gd
daniel-k-taylor: That should be fine, I just want to make sure we're not adding code around needlessly.

scenes/game/game.gd
Mach131: Based on cpat's videos, I was thinking that the issues come up when the user and the game logic try to open/close a popup at the same time (e.g. clicking the gauge window when it's about to bring it up to ask for a force/gauge cost); my computer's too fast to reproduce the same behavior, but maybe it would be smarter to just not open a new popup if one's already open? Would probably need to combine with that indicator of the zone it wants you to click, though.

I don't seem to be able to run the game in browser mode, it says something about a missing template file?

globals/card_definitions.gd
daniel-k-taylor: i'll let you decide if single or double looks better. We just don't currently use that anywhere in the UI, which maybe is what you're looking for.

scenes/game/game.gd
```diff
@@ -1150,6 +1150,15 @@ func _stat_notice_event(event):
 	spawn_damage_popup(notice_text, player)
 	return SmallNoticeDelay
 
+func _set_card_bonus(card_id, bonus, value=true):
+	var card = find_card_on_board(card_id)
+	if bonus == "ex":
+		card.set_ex(value)
+	if bonus == "wild":
```
daniel-k-taylor, line 1157: kind of weird to have 2 ifs and one elif.
I mean, it won't matter, but just odd.

daniel-k-taylor (approved):

PR #98, opened 2024-02-17
Cancellable Character Actions

Mach131:
- Made it so that some UAs can be cancelled before you actually make the first decision
- Of those, made some UAs accessible through the "shortcut menu" when you select cards while picking an action
I generally avoided applying this to UAs with a force/gauge cost, since the ordering would get weird and you can cancel those already anyway

Strike UAs (in shortcut)
- Chaos (default)
- Gio (default)
- Londrekia (default)
- Nanase (default)
- Phonon (default)
- Seth (default)

Gauge-From-Hand UAs (in shortcut)
- S6 Revert
- Yuzuriha (default)

Boost from Gauge UAs (in shortcut)
- Linne (exceed)

Choose Discard UAs (in shortcut)
- MIka (default)

UAs beginning with a choice
- Hyde (default)
- Litchi (summon/remove mantenbo)
- Propeller Knight (exceed)
- Rachel
- Shovel/Shield Knight (default)

Buddy-placement UAs
- Carl
- Litchi (move mantenbo)
- Mole Knight
- Shovel/Shield Knight (exceed)

### DIFF
--- a/scenes/game/game.gd
+++ b/scenes/game/game.gd
@@ -1386,13 +1386,20 @@ func _on_boost_played(event):
 	spawn_damage_popup("Boost!", player)
 	return BoostDelay
 
-func _on_choose_card_hand_to_gauge(event):
+func _on_choose_card_hand_to_gauge(event, can_cancel=false):
 	var player = event['event_player']
 	var min_amount = event['number']
 	var max_amount = event['extra_info']
 	select_card_destination = game_wrapper.get_decision_info().destination
 	if player == Enums.PlayerId.PlayerId_Player and not observer_mode:
-		begin_discard_cards_selection(min_amount, max_amount, UISubState.UISubState_SelectCards_DiscardCardsToGauge)
+		if prepared_character_action_data and prepared_character_action_data['effect_type'] == 'gauge_from_hand' and not preparing_character_action:
+			var selected_card_ids = prepared_character_action_data['hand_to_gauge_cards']
+			var success = game_wrapper.submit_card_from_hand_to_gauge(Enums.PlayerId.PlayerId_Player, selected_card_ids)
+			if success:
+				prepared_character_action_data = {}
+				change_ui_state(UIState.UIState_WaitForGameServer)
+		else:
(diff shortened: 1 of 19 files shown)


scenes/game/game.gd
```diff
@@ -3046,6 +3100,85 @@ func _update_buttons():
...
 			button_choices.append({ "text": strike_text, "action": _on_shortcut_strike_pressed, "disabled": not can_strike or not game_wrapper.can_do_strike(Enums.PlayerId.PlayerId_Player) })
 			button_choices.append({ "text": boost_text, "action": _on_shortcut_boost_pressed, "disabled": not can_boost or not game_wrapper.can_do_boost(Enums.PlayerId.PlayerId_Player) })
+
+			# Check for character actions with card-related shortcuts
+			for i in range(game_wrapper.get_player_character_action_count(Enums.PlayerId.PlayerId_Player)):
+				var char_action = game_wrapper.get_player_character_action(Enums.PlayerId.PlayerId_Player, i)
+				if 'shortcut_effect_type' in char_action:
+					if char_action['shortcut_effect_type'] == "strike":
+						var action_possible = game_wrapper.can_do_character_action(Enums.PlayerId.PlayerId_Player, i)
+						var action_name = "Strike with Character Action"
```
daniel-k-taylor, line 3110: feels like there is a ton of duplicated code between each of these ifs

daniel-k-taylor: Can you just have a table to map shortcut effect to action name and string or something?
You'll still have to set up like mincards and stuff in the ifs, but you can move a lot out that way.

daniel-k-taylor: things that can be outside the big if block:
action possible
get action name from a table or char_action['action_name'] or default to 'Character Action'
reading force_cost/gauge_cost

daniel-k-taylor (changes requested):

scenes/game/game.gd
Mach131: Didn't quite set up a table because of the conditions, but let me know if it's a bit cleaner now

daniel-k-taylor (approved):

PR #100, opened 2024-02-18
Match Reporting

daniel-k-taylor:
Call the new server API to report the match when completed.
Also change client version in debug so we can ignore those matches.


scenes/game/game.gd
```diff
@@ -1863,6 +1865,7 @@ func _on_game_over(event):
 		game_over_label.text = "DEFEAT"
 	else:
 		game_over_label.text = "WIN!"
+		game_wrapper.submit_match_result()
```
Mach131, line 1868: I guess the one thing that might confound data a bit is if people disconnect/dodge, though there's not really a good way around that since they might leave when they're about to lose anyway

Mach131 (approved):

PR #106, opened 2024-02-19
Add a game timer - currently not enforced

daniel-k-taylor:


Mach131 (dismissed):
Looks good to me; it seems like it's hidden in AI games? Which is what we'd want, but can't judge how it looks until it goes out

PR #112, opened 2024-02-25
Rework card popout behavior

daniel-k-taylor:


scenes/game/game.gd
```diff
@@ -4424,152 +4401,55 @@ func card_in_selected_cards(card):
...
 
 func close_popout():
-	card_popout.visible = false
-	if popout_showing_node:
-		popout_showing_node.z_index = 0
-	await clear_card_popout()
+	while card_popout_parent.get_child_count() > 0:
+		var child = card_popout_parent.get_child(0)
+		card_popout_parent.remove_child(child)
+		child.queue_free()
```
Mach131, line 4425: From a quick search of what queue_free does, it seems like the remove_child call might be redundant; have you tried seeing if anything changes if you delete that call?

Mach131 (approved):
Tried checking out the branch and it seems to work smoothly for me! Again, my computer was generally good enough to avoid issues before, but the popups definitely seem to close more cleanly now

scenes/game/game.gd
daniel-k-taylor: even though queue_free does remove the node from the scene tree, I want to be super explicit about removing it immediately here because i'm also adding a child next and referring to child 0, so I just want to be very sure I'm referring to the new one and not the old one.

PR #113, opened 2024-02-25
King knight

Mach131:


scenes/game/local_game.gd
```diff
@@ -9577,7 +9618,10 @@ func do_force_for_effect(performing_player : Player, card_ids : Array, treat_ult
 		var effect_times = 0
 		if decision_info.effect['per_force_effect']:
 			decision_effect = decision_info.effect['per_force_effect']
-			effect_times = force_generated
+			var interval = 1
+			if 'force_effect_interval' in decision_info.effect:
+				interval = decision_info.effect['force_effect_interval']
+			effect_times = floor(force_generated / interval)
```
daniel-k-taylor, line 9624: curious if there is a warning here about integer division.
I'm guessing maybe not since it can't know for sure interval is an integer.
If there is, just make interval = 1.0

assets/cardbacks/cardback_king_decree.png.import
```diff
@@ -0,0 +1,34 @@
+[remap]
+
+importer="texture"
```
daniel-k-taylor, line 3: this file isn't used?

daniel-k-taylor (changes requested):

scenes/game/local_game.gd
Mach131: no warning, but changed just in case

daniel-k-taylor (approved):

PR #114, opened 2024-02-27
Chunli

daniel-k-taylor:


scenes/game/local_game.gd
```diff
@@ -6914,6 +6947,14 @@ func get_card_stat(check_player : Player, card : GameCard, stat : String) -> int
...
 	elif str(value) == "RANGE_TO_OPPONENT":
 		value = check_player.distance_to_opponent()
+	elif str(value) == "INCLUDE_OPPONENT_IF_MOVED_PAST":
+		value = -1
+		# NOTE: This probably isn't necessary for any mechanics as the card that uses this
+		# does the effect range_includes_if_moved_past.
+		# However, this will be slightly wrong if someone has a +range boost of any kind.
+		# If a character can do that and also cares about range, then worry about that then.
+		if active_strike and check_player.strike_stat_boosts.range_includes_opponent:
+			value = check_player.distance_to_opponent()
```
Mach131, line 6957: Does this interact properly with range_parry? i.e. if gief names 3 when you're at distance 3 from the opponent, this probably shouldn't be a valid target. I forget exactly how that's handled but might be an appropriate time to leave it as -1 (if that still gets processed properly by range checking)?

Mach131: Actually ignore this, since this only happens during active_strike i don't think we'll ever have to worry about range parries

Mach131 (approved):

PR #115, opened 2024-02-28
Vega

daniel-k-taylor:


data/card_definitions.json
```diff
@@ -22915,6 +22915,363 @@
...
+	},
+	{
+		"id": "vega_scarletterror",
+		"type": "special",
+		"display_name": "Scarlet Terror",
+		"force_cost": 0,
+		"gauge_cost": 0,
+		"range_min": 1,
+		"range_max": 2,
+		"speed": 6,
```
Mach131, line 23036: Just checking, reddit DB says this is speed 4?

data/card_definitions.json
```diff
@@ -22915,6 +22915,363 @@
...
+		"boost": {
+			"boost_type": "immediate",
+			"force_cost": 2,
+			"cancel_cost": -1,
+			"display_name": "Repel Leap",
+			"effects": [
+				{
+					"timing": "immediate",
+					"condition": "at_edge_of_arena",
+					"effect_type": "vega_ua_guarantee",
```
Mach131, line 23062: I wonder if there's a way to implement this (and the regular UA) with character_bonus instead? probably not a big deal though

data/decks/vega.json
```diff
@@ -0,0 +1,209 @@
...
+				"character_effect": true
+			},
+			"gauge_max": 1
+		},
+		{
+			"timing": "set_strike",
+			"condition": "at_edge_of_arena",
+			"effect_type": "add_attack_effect",
+			"added_effect": {
+				"condition": "not_has_vega_ua_guarantee",
```
Mach131, line 24: yeah maybe both this and the boost could set character_bonus, then the set_strike effect just checks that (avoids a bit of repetition)

data/card_definitions.json
```diff
@@ -22915,6 +22915,363 @@
...
+						{ "effect_type": "advance", "amount": 3, "bonus_effect": {
+							"condition": "advanced_through",
+							"effect_type": "choose_discard",
+							"limitation": "continuous",
+							"destination": "play_boost",
+							"amount": 1,
+							"amount_min": 1,
+							"discard_effect": {
+								"effect_type": "boost_then_sustain_topdiscard",
+								"amount": "DISCARDED_COUNT",
```
Mach131, line 23160: I think we do already have a play_boost effect or something similar (one of propeller knight's boosts plays another boost from discard), in case that makes this any easier to implement

Mach131 (dismissed):

data/card_definitions.json
daniel-k-taylor: that's correct. dunno where this 6 came from :)

data/card_definitions.json
daniel-k-taylor: ah, good point, i'll switch to that i guess.

data/decks/vega.json
daniel-k-taylor: okay, yea, i like that better.

Mach131 (approved):

PR #128, opened 2024-03-14
Merkava

Mach131:


data/card_definitions.json
```diff
@@ -15061,6 +15061,397 @@
...
+				"condition_detail": "I, Persistently Cling",
+				"effect_type": "reveal_strike",
+				"and": {
+					"effect_type": "calculate_range_from_buddy"
+				}
+			},
+			{
+				"timing": "after",
+				"condition": "not_hit_opponent",
+				"effect_type": "return_attack_to_hand",
```
daniel-k-taylor, line 15420: Does cleanup still happen in this case? If not, need to remove the buddy here too

daniel-k-taylor (approved):

data/card_definitions.json
Mach131: Yup, active_strike still remembers the card when it gets removed

PR #135, opened 2024-04-04
Refactor combinatorial functions

rngesus-wept:
This changes the generation approach from recursive to iterative, and makes the functions in question return results directly instead of by side-effecting one of the input arguments.

Tested: GUT. In addition, GUT runtime is reduced from ~44 seconds to ~17 seconds, a ~2.5x improvement. (At least, excluding the time the program takes to load up in the first place.)

### DIFF
--- a/scenes/game/ai_player.gd
+++ b/scenes/game/ai_player.gd
@@ -335,6 +335,10 @@ func get_move_actions(game_logic : LocalGame, me : LocalGame.Player, opponent : 
 				possible_move_actions.append(MoveAction.new(i, combo, used_free_force))
 	return possible_move_actions
 
+## Given a list of card IDs `cards`, return all combinations of card IDs that can be spent to generate
+## `force_target` Force, with up to `free_force_available` units worth of slack. The actual return value
+## is a list of pairs [combination, free_force_used], where the latter is true iff at least one point of
+## the "slack" is necessary for the combination.
 func generate_force_combinations(game_logic : LocalGame, me : LocalGame.Player, cards, force_target, free_force_available):
 	var current_force = me.force_cost_reduction
 	var card_db = game_logic.get_card_database()
--- a/scenes/game/ai_player.gd
+++ b/scenes/game/ai_player.gd
@@ -350,8 +350,8 @@ func generate_force_combinations(game_logic : LocalGame, me : LocalGame.Player, 
 	for card_id in cards:
 		var card_force_value = card_db.get_card_force_value(card_id)
 		for i in range(candidates.size()):
-			# Set iteration range prior to adding anything or you'll end up reprocessing new candidates
-			# that were already added during the current pass.
+			# Only process candidates that already existed at the beginning of this iteration
+			# (that is, ignore the ones we're about to append below, to avoid duplication).
 			var combination = candidates[i]
 			if combination[0] < force_target:
 				if combination.size() == 1:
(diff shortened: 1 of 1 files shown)


scenes/game/ai_player.gd
```diff
@@ -328,40 +328,61 @@ func get_move_actions(game_logic : LocalGame, me : LocalGame.Player, opponent :
...
-
-	for i in range(current_index, cards.size()):
-		current_combination.append(cards[i])
-		generate_card_count_combinations(cards, hand_size, current_combination, i + 1, combinations)
-		current_combination.pop_back()
+	if current_force == force_target:
+		return [[[], false]]
+
+	var candidates = [[current_force]]  # Each entry in result is a list whose car is a force count and whose cdr is a list of cards
+	for card_id in cards:
```
daniel-k-taylor, line 349: How does this not just fill up the candidates with the first X cards for your force target of X?
I'm not quite following this algorithm.

daniel-k-taylor: Oh, it only appends more candidates, it doesn't add to existing candidates.
I think I get it now.

daniel-k-taylor (changes requested):
I think I get it now, but can you try to explain it better in comments? I had to look at this for a little while to grasp what was going on.

scenes/game/ai_player.gd
```diff
@@ -328,40 +328,61 @@ func get_move_actions(game_logic : LocalGame, me : LocalGame.Player, opponent :
...
-		return
-
-	for i in range(current_index, cards.size()):
-		current_combination.append(cards[i])
-		generate_card_count_combinations(cards, hand_size, current_combination, i + 1, combinations)
-		current_combination.pop_back()
+	if current_force == force_target:
+		return [[[], false]]
+
+	var candidates = [[current_force]]  # Each entry in result is a list whose car is a force count and whose cdr is a list of cards
```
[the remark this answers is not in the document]
rngesus-wept: Changed to "head" and "remainder".

scenes/game/ai_player.gd
rngesus-wept: Yeah; we add candidates because when we're thinking about a particular card X we want all subsets that can contain X, but we still want to have all the subsets that don't contain X in our list because we might (duplicate and) add other stuff to them later.

rngesus-wept:
Updated comment phrasing to be less arcane.

scenes/game/ai_player.gd
```diff
@@ -328,40 +328,62 @@ func get_move_actions(game_logic : LocalGame, me : LocalGame.Player, opponent :
...
-		generate_card_count_combinations(cards, hand_size, current_combination, i + 1, combinations)
-		current_combination.pop_back()
+		return [[[], false]]
+
+	var candidates = [[current_force]]  # Each entry in result is a list whose first element is a force count
+	                                    # and whose remainder is a list of cards
+	for card_id in cards:
+		var card_force_value = card_db.get_card_force_value(card_id)
+		for i in range(candidates.size()):
+			# Only process candidates that already existed at the beginning of this iteration
```
daniel-k-taylor, line 353: now that I understand this more, this works because the inside of range() is only evaluated once at the start?

daniel-k-taylor (approved):

scenes/game/ai_player.gd
rngesus-wept: That's right.

PR #136, opened 2024-04-06
Pull common unit testing functions to a parent class and refactor

rngesus-wept:
A bunch of functions common to all the test classes, including the oft-used `execute_strike`, are moved to a parent class `ExceedGutTest` and then made available to unit test classes through inheritance. In addition, the following refactoring changes were made:

  * All player decisions are collapsed to a single array. It is up to the test author to structure this input so that its order and content types match the order and content types of the decisions that will be offered. When converting an old test, this means that some decisions will need to be specified when they didn't used to be. The original code would default to "choice 0" if you didn't say anything, but the new code will generate a failure indicating that it wants instructions. Error/failure messages are structured to hopefully provide context to make it easy to figure out where the test got confused.

    Why do it this way? While beginning to merge tests to share common behavior, I noticed that the function signature of `execute_strike` would vary based on the new needs of a character. Aside from the more common attack effect decisions, some characters need to try paying Gauge costs in particular ways, or setting attacks face up, and so on and so on. Over time, this causes the argument list of `execute_strike` to bloat, and tests start having to add a bunch of "blank" arguments (like `[]` or `0` or `false`, depending on type) in order to get to the part of the argument list it actually cares about. (This is partially a fault of gdscript itself for not supporting keyword arguments.) For a unified test suite, this is a huge maintenance burden.

    So from now on, if you want to support a *new* kind of decision, all you have to do is add it in the right place in the list. Instead of adding an argument to support new behavior, simply add a branch of the appropriate enum to one of the `match` statements in `process_decisions` or `process_remaining_decisions` in `exceed_test.gd`. Error messages are provided to help indicate when a decision type needs more support.
    
  * Test card IDs are automatically assigned by a counter that just spits out a fresh ID any time you ask for one. This lowers the risk of using a duplicate ID (in which case the game engine will complain about you trying to manipulate a component that has already been `free`d). (For example, if you needed to `execute_strike` twice in one test, the old code would try to reuse `TestCardId1` and `TestCardId2` the second time around.) You can track as many cards as you need to for any test case without having to add new `TestCardIdN` constants. You will need to assign a variable to track the id number whenever you need to use it in more than one place, but hopefully this also makes the code more readable since you can talk about `assault_id` as clearly being an Assault card instead of remembering what kind of thing `TestCardId3` points to.

    The upshot: `give_specific_cards`, `give_player_specific_card`, `give_gauge`, and `execute_strike` all return integers or lists of integers: newly generated id(s) for later use in the test. Remember to bind the return values to something when it's relevant.

    Second upshot: `do_and_validate_strike` and `execute_strike` no longer return the event queue. In order to preserve the event queue so that tests that care about it can inspect it without surprises (i.e. they get a clean and complete copy of the event queue when they ask for it), these functions now do not interact with the event queue at all, and the validation check in `do_and_validate_strike` is solely based on the game state.

  * Code that uses `fail_test` to signal a bad input value or similar also try to `return` immediately afterwards. This is because `fail_test` does not actually abort the test, so if the code continues running and encounters an error, the test runner throws an error into the **editor**... which stops test suite execution and clobbers any useful test failure messages that would have been presented upon completion.

    (Doing this in a more comprehesively correct way is stymied by the lack of proper exception handling in gdscript.)

  * Some core libraries have gained a `class_name`, which makes them accessible globally without having to `const MyLibrary = preload(...)` and allows for inheritance between script files. There is no disruption to unconverted code; it's still okay to `preload` those files into a `const` even if it won't be necessary anymore. (Downside: Every time you mess with `class_name`s, the Godot editor will insist on a project reload so it can do Magical IDE Things (tm).)

The test suites for Waldstein, Yuzuriha, and Zangief have been converted. There is no disruption to other test code; the other existing tests continue to run using the original framework and will continue to do so until they're converted to `extend ExceedGutTest`.

### DIFF
--- a/scenes/game/enums.gd
+++ b/scenes/game/enums.gd
@@ -1,3 +1,4 @@
+class_name Enums
 extends Node
 
 enum PlayerId {
@@ -21,17 +22,17 @@ enum DecisionType {
 	DecisionType_BoostNow,
 	DecisionType_ChooseArenaLocationForEffect,
 	DecisionType_ChooseDiscardContinuousBoost,
-	DecisionType_ChooseDiscardOpponentGauge,
+	DecisionType_ChooseDiscardOpponentGauge,  # 5
 	DecisionType_ChooseFromBoosts,
 	DecisionType_ChooseFromDiscard,
 	DecisionType_ChooseFromTopDeck,
 	DecisionType_ChooseSimultaneousEffect,
-	DecisionType_EffectChoice,
+	DecisionType_EffectChoice,  # 10
 	DecisionType_ForceBoostSustainTopdeck,
 	DecisionType_ForceBoostSustainTopDiscard,
 	DecisionType_ForceForEffect,
 	DecisionType_GaugeForEffect,
-	DecisionType_NameCard_OpponentDiscards,
+	DecisionType_NameCard_OpponentDiscards,  # 15
 	DecisionType_ChooseToDiscard,
 	DecisionType_PayStrikeCost_Required,
 	DecisionType_PayStrikeCost_CanWild,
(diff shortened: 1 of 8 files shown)


scenes/game/enums.gd
```diff
@@ -72,102 +73,102 @@ enum EventType {
...
 	EventType_Strike_SetX,
 	EventType_Strike_Started,
-	EventType_Strike_Started_Ex,
+	EventType_Strike_Started_Ex,     # 95
 	EventType_Strike_Started_ExtraAttack,
 	EventType_Strike_Stun,
 	EventType_Strike_Stun_Immunity,
 	EventType_Strike_TookDamage,
-	EventType_Strike_WildStrike,
+	EventType_Strike_WildStrike,     # 100
```
daniel-k-taylor, line 171: I initially had something like these number comments but found it too annoying to keep updating them as I also wanted to have the list be alpha sorted.

scenes/game/local_game.gd
```diff
@@ -43,7 +44,7 @@ const StrikeStaticConditions = [
 	"canceled_this_turn",
 ]
 
-var event_queue = []
+var event_queue = []  # This is purely a logging utility
```
daniel-k-taylor, line 47: no it isn't. This is how the UI functions.

test/exceed_test.gd
```diff
@@ -0,0 +1,279 @@
...
+var game_logic : LocalGame
+var next_test_card_id = 50000
+
+var player1 : LocalGame.Player
+var player2 : LocalGame.Player
+
+### !!! IMPORTANT
+# Subclasses must override this function
+func who_am_i():
+	return "no-one-lmao"
```
daniel-k-taylor, line 13: maybe a descriptive string here on what you intend this to be in the overridden func

test/exceed_test.gd
```diff
@@ -0,0 +1,279 @@
...
+							"%s failed to discard cards %s for armor" % [player, choice])
+				var decision_type:  # Unknown decision type, just roll with it?
+					if typeof(choice) == Variant.Type.TYPE_ARRAY:
+						fail_test("Attempting to apply array choice %s to a decision of type %s" % [
+								choice, Enums.DecisionType.keys()[decision_type]])
+						return
+					assert_true(game_logic.do_choice(player, choice),
+							"Decision of type %s unhandled by test harness (attempted by player %s, content %s)" % [
+									decision_type, player, choice])
+		## TODO: Does the loop need to wait a tick or two here for the game engine to
```
daniel-k-taylor, line 218: there is nothing time based inside localgame

test/exceed_test.gd
```diff
@@ -0,0 +1,279 @@
...
+						"Decision of type %s unhandled by test harness (attempted by player %s, content %s)" % [
+								decision_type, player, content])
+
+func process_remaining_decisions(initiator, defender, init_choices, def_choices):
+	var empty_loop_count = 0
+	while init_choices.size() + def_choices.size() >= 1:
+		# empty_loop_count is used to detect when there are still prescribed choices in
+		# the input, but the game is not actually making additional decision points available.
+		empty_loop_count += 1
+		if empty_loop_count >= 3:
```
daniel-k-taylor, line 182: I don't get it, why 3?

test/exceed_test.gd
```diff
@@ -0,0 +1,279 @@
...
+			fail_test("Game is not providing decision points to process initiator" +
+					" choices %s or defender choices %s" % [init_choices, def_choices])
+			return
+		while game_logic.game_state == Enums.GameState.GameState_PlayerDecision:
+				## TODO: Figure out if it's really necessary to limit ourselves to this decision type
```
daniel-k-taylor, line 187: what do you mean by "this decision type"? the game state?

test/exceed_test.gd
```diff
@@ -0,0 +1,279 @@
...
+				Enums.DecisionType.DecisionType_ChooseSimultaneousEffect:
+					assert_true(game_logic.do_choice(player, choice),
+							"%s failed to perform a choice with value %s" % [player, choice])
+				Enums.DecisionType.DecisionType_ChooseToDiscard:
+					assert_true(game_logic.do_choose_to_discard(player, choice),
+							"%s failed to discard cards %s" % [player, choice])
+				Enums.DecisionType.DecisionType_ForceForArmor:
+					assert_true(game_logic.do_force_for_armor(player, choice),
+							"%s failed to discard cards %s for armor" % [player, choice])
+				var decision_type:  # Unknown decision type, just roll with it?
```
daniel-k-taylor, line 210: I think there is a huge handler for all the different types in the randomai test. Not sure how relevant it is here.

test/exceed_test.gd
```diff
@@ -0,0 +1,279 @@
...
+		fail_test("Insufficient decisions defined for player %s during strike" % game_logic.decision_info.player)
+		return
+
+func execute_strike(initiator, defender, init_card: String, def_card: String,
+		init_ex = false, def_ex = false, init_choices = [], def_choices = []):
+	var init_card_id = -1
+	var init_card_ex_id = -1
+	var def_card_id = -1
+	var def_card_ex_id = -1
+	## TODO: Because all the card IDs are generated internally and at runtime, it is hard
```
daniel-k-taylor, line 231: what is there todo here?

test/exceed_test.gd
```diff
@@ -0,0 +1,279 @@
...
+	##   to pass those IDs in through *_choices
+	if init_card:
+		init_card_id = give_player_specific_card(initiator, init_card)
+		if init_ex:
+			init_card_ex_id = give_player_specific_card(initiator, init_card)
+		do_and_validate_strike(initiator, init_card_id, init_card_ex_id)
+	else:
+		init_card_id = do_and_validate_strike(initiator, -1)  # wild swing
+	## TODO: Why no all_events modification in the initiator block? Is it
```
daniel-k-taylor, line 240: Not really sure what this is asking

daniel-k-taylor (changes requested):

scenes/game/local_game.gd
rngesus-wept: Ah, I see my mistake. I was looking for code that reads `event_queue` but forgot that most code will access it through `get_latest_events()`.

scenes/game/enums.gd
rngesus-wept: Got it, reverted.

test/exceed_test.gd
rngesus-wept: See below.

test/exceed_test.gd
rngesus-wept: If a test writer has composed the test incorrectly to specify more decisions than the game actually makes available, the choices.size() check will never fail and will gets stuck in this loop. The `empty_loop_count` check is there to break out of the loop after enough time[1] goes by without being presented with another game decision.

The actual number in this line is just a number of retries. Any small number would be fine, depending on your tolerance for the game advancing in parallel with the test[1].

[1] Some of this is based on the assumption that the game code runs in parallel with the test code, so it is possible for there to be non-negligible clock time between two consecutive decisions if a bunch of other Game Stuff has to happen in between. In those cases we don't want to jump out of the decision-resolving loop too fast. This is also the significance of the TODO (and the commented `wait_seconds`) for the previous comment -- the speed of iteration has to be tuned to some sense of what a tolerable amount of clock time between decisions is.

So then there are two cases to consider when you have leftover decision specs and no decision is being presented: either the test (or the code itself) is incorrectly written, and you should flag a failure; or the game code is just slow, in which case you should just wait a bit for the next decision and give it a chance to pass. The iteration speed and the number of retries in the `empty_loop_count` check reflect the point at which you decide you're in the first case instead of the second.

test/exceed_test.gd
rngesus-wept: Ah, yeah, the comment should say game state.

The comment as a whole is there because this function replaces `handle_simultaneous_decisions` (see any of the unmodified test files), which specifies that it only cares about that that game state + `DecisionType_ChooseSimultaneousEffect`, and I'm not sure how many or which sorts of things fall under the umbrella of that decision type as opposed to one of the others.

test/exceed_test.gd
rngesus-wept: Oh interesting. It looks like a lot of that stuff assumes that there's an AI picking certain options, but it does give the correct mapping between decision types and `local_game.do_*` functions. I can incorporate that in a future push request.

test/exceed_test.gd
rngesus-wept: Rephrased to: 
```python
	## TODO: Figure out what to do if one of the *_choices needs to use a card
	##   ID that will not be assigned until the `give_player_specific_card`
	##   calls below.
```

test/exceed_test.gd
rngesus-wept: The original behavior of `execute_strike` includes a return of `all_events`, which is the aggregate contents of `local_game.event_queue` over the duration of `execute_strike`. This return value included all events that happened during the defender's response and the decision-making that ensues, but omits the portion of the event queue that happens during the initiator's attack set. The initiator's attack set calls through `do_and_validate_strike`, which inspects `event_queue` (and therefore *empties* it) without returning its contents or adding a copy to `all_events`, so those events are missing. This might be surprising to anyone who looked at the array of events returned by the original `execute_strike` and assumes that it is a *complete* array of events from the strike.

It was unclear to me whether it was important that any manipulation of `all_events` happens during `execute_strike`. I chose to modify things to leave `get_latest_events()` alone entirely for all of `execute_strike`. Any test that actually cares about its contents can call `get_latest_events` directly and get a complete array of relevant events.

test/exceed_test.gd
rngesus-wept: Done.

daniel-k-taylor (approved):

test/exceed_test.gd
daniel-k-taylor: I'm going to go with - i probably didn't think about it too much and execute_strike grew organically and I only did what I needed for the tests I wrote.

PR #138, opened 2024-04-08
Use a dictionary instead of an array for card database

rngesus-wept:
This changes the backing data structure of the card and deck databases from an Array to a Dictionary. The upshot is that all (well, most) operations on either database -- adding, looking up, and deleting cards -- now take constant time instead of linear time.

There is one functional change: `get_card(id)` now returns `null` for all unrecognized card IDs. The original behavior was that it returns `"(NO CARD)"` for id -1 and `null` for all *other* unrecognized card IDs. There were no tests covering this, and nothing obvious broke, so I assume there isn't anything that relies on `get_card` sometimes returning a String instead of a GameCard or `null`. But maybe the difference is important somehow.

There does not appear to be any code outside of this file that accesses CardDatabase.all_cards directly, and the tests all pass.

Tested: Full test time is reduced from ~18s to ~16s, about a 10% improvement.

### DIFF
--- a/globals/card_definitions.gd
+++ b/globals/card_definitions.gd
@@ -4,33 +4,25 @@ var card_data = []
 
 var card_definitions_path = "res://data/card_definitions.json"
 var decks_path = "res://data/decks"
-var decks = []  # An array of (JSON) dictionaries
+var decks = {}  # A dictionary of (JSON) dictionaries
 
 const CardHighlightColor = "#7DF9FF" # Light blue
 
 func get_deck_test_deck():
-	for deck in decks:
-		if deck['id'] == "rachel":
-			return deck
-	return get_random_deck(-1)
+	return decks.get("rachel", get_random_deck(-1))
 
 func get_random_deck(season : int):
 	# Randomize
 	if season == -1:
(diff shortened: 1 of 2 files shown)


globals/card_definitions.gd
```diff
@@ -4,33 +4,25 @@ var card_data = []
...
-		var season_decks = []
-		for deck in decks:
-			if deck['id'] in GlobalSettings.CharacterBanlist:
-				continue
-			if deck['season'] == season:
-				season_decks.append(deck)
-		var random_index = randi() % len(season_decks)
-		return season_decks[random_index]
-
+		var season_decks = decks.values().filter(func (deck):
```
daniel-k-taylor, line 23: It probably makes more sense to do this above to filter out the banlist in a similar way for the -1 season.

daniel-k-taylor (approved):
I doubt anything relied on get_card returning a string.

globals/card_definitions.gd
rngesus-wept: Good call. Fixed and tested.

daniel-k-taylor (approved):

PR #142, opened 2024-04-19
Rename do_card_from_hand_to_gauge

rngesus-wept:
It moves the card from hand to *wherever the current decision wants them to go*.

There are no more hits for `do_card_from_hand_to_gauge` in the codebase, and all the unit tests pass, so this should be complete and correct.

### DIFF
--- a/scenes/game/game_wrapper.gd
+++ b/scenes/game/game_wrapper.gd
@@ -441,7 +441,7 @@ func submit_discard_to_max(player : Enums.PlayerId, card_ids : Array) -> bool:
 	var game_player = _get_player(player)
 	return current_game.do_discard_to_max(game_player, card_ids)
 
-func submit_card_from_hand_to_gauge(player : Enums.PlayerId, card_ids : Array) -> bool:
+func submit_relocate_card_from_hand(player : Enums.PlayerId, card_ids : Array) -> bool:
 	var game_player = _get_player(player)
 	return current_game.do_relocate_card_from_hand(game_player, card_ids)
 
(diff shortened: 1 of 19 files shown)


scenes/game/remote_game.gd
```diff
@@ -233,7 +233,7 @@ func do_card_from_hand_to_gauge(player : LocalGame.Player, card_ids : Array) ->
 func process_card_from_hand_to_gauge(action_message) -> void:
```
daniel-k-taylor, line 233: you didn't rename this

scenes/game/game_wrapper.gd
```diff
@@ -442,7 +442,7 @@ func submit_discard_to_max(player : Enums.PlayerId, card_ids : Array) -> bool:
 
 func submit_card_from_hand_to_gauge(player : Enums.PlayerId, card_ids : Array) -> bool:
```
daniel-k-taylor, line 443: you didn't rename this

scenes/game/remote_game.gd
```diff
@@ -221,7 +221,7 @@ func process_discard_to_max(action_message) -> void:
 	var card_ids = action_message['card_ids']
 	local_game.do_discard_to_max(game_player, card_ids)
 
-func do_card_from_hand_to_gauge(player : LocalGame.Player, card_ids : Array) -> bool:
+func do_relocate_card_from_hand(player : LocalGame.Player, card_ids : Array) -> bool:
 	var action_message = {
 		'action_type': 'action_card_from_hand_to_gauge',
```
daniel-k-taylor, line 226: Ideally we would update this too. But do we care about replay compatibility?
I'd say probably not at this point?

Actually, isn't this how these functions are called? You renaming that is going to break it.

daniel-k-taylor (changes requested):
I think this breaks remote play.
You need to rename the action too.

Please look at _process_game_message and make sure this works.
Make sure you test playing with yourself with 2 clients and trigger this effect.

rngesus-wept:
Updated, PTAL

daniel-k-taylor (approved):

PR #147, opened 2024-05-07
Flesh out AI-internal objects for representing game state

rngesus-wept:
AI(Game|Player|Strike)State are now Resources, so that it's easy to make manipulable copies. This is with an eye toward writing AI functionality that works with hypothetical scenarios.

The top level AIPlayer object now holds references to the actual game and player objects, so that most calls should refer to those objects directly instead of needing them passed in all the time. It's a bit of a stopgap -- what I'd really like most of the time is to be able to pass in a AIGameState and have functionality refer to that.

Unit tests pass, played through a full AI game or ten both within the Godot client and within a local HTML5 deployment, because apparently there are subtle differences in engine behavior. :neutral_face:

### DIFF
--- a/scenes/game/ai_player.gd
+++ b/scenes/game/ai_player.gd
@@ -28,7 +28,6 @@
 ##         the top of a turn; for example, a mid-strike decision on movement.
 
 class_name AIPlayer
-extends Node2D
 
 const TEST_PrepareOnly = false
 
@@ -45,7 +44,7 @@ var ai_policy
 var game_player: LocalGame.Player
 var game_opponent: LocalGame.Player
 
-func initialize(local_game: LocalGame, player: LocalGame.Player, policy = null):
+func _init(local_game: LocalGame, player: LocalGame.Player, policy = null):
 	game_logic = local_game
 	game_player = player
 	game_opponent = local_game._get_player(local_game.get_other_player(player.my_id))
@@ -126,7 +125,7 @@ class AIPlayerState:
 			if name in AIPlayer.IGNORE_PROPERTIES or name == 'source':
 				continue
 
-			var value = self._get(name)
+			var value = self.get(name)
 			if value == null:
 				new_state._set(name, null)
 				continue
@@ -172,7 +171,7 @@ class AIStrikeState:
 			if name in AIPlayer.IGNORE_PROPERTIES or name == 'source':
 				continue
 
-			var value = self._get(name)
+			var value = self.get(name)
 			if value == null:
 				new_state._set(name, null)
 				continue
@@ -230,7 +229,7 @@ class AIGameState:
 			if name in AIPlayer.IGNORE_PROPERTIES or name == 'source':
 				continue
 
-			var value = self._get(name)
+			var value = self.get(name)
 			if value == null:
 				new_state._set(name, null)
 				continue
--- a/scenes/game/ai_player.gd
+++ b/scenes/game/ai_player.gd
@@ -1,4 +1,4 @@
-## Basic framework for AI behavior.
+## AI player controller object.
 
 ## This file describes a bunch of hooks and basic functions used to provide an
 ## AI with options, then relay the AI's choice from those options back to the
@@ -6,16 +6,18 @@
 ## [i]policy[/i], which is a Node that implements the laundry list of pick_*
 ## functions. See ai_random_policy.gd for a very basic (but complete!) example.
 
-## After initialization, the main entry point into the code is through
-## [method take_turn], which is called by game.gd's `ai_take_turn`. `take_turn`
-## accepts a summary of the game state and returns an instance of one of the
-## various *Action classes defined in this file. `ai_take_turn` then uses
-## that return value to manifest the appropriate changes in the game proper.
+## The player is initialized with a reference to the relevant game and player
+## objects. After that, the main entry point into the code is through [method take_turn],
+## which is called by game.gd's `ai_take_turn`. `take_turn` consults the stored
+## game state the game state and returns an instance of one of the various
+## *Action classes defined in this file. `ai_take_turn` then uses that return
+## value to manifest the appropriate changes in the game proper.
 
-## `take_turn` enumerates all legal actions for the AI to take, then passes
-## them to policy and handles the return. Note that each possible variant of
-## a (game) action counts as a distinct (AI) action; for example, the choice
-## of which cards to discard for Force generation during a Walk.
+## `take_turn` enumerates all legal actions for the AI to take, then passes them
+## to policy (and then forwards the return value back to game.gd as described
+## above). Note that each possible variant of a (game) action counts as a
+## distinct (AI) action; for example, the choice of which cards to discard for
+## Force generation during a Walk.
 
 ## A broad overview of the main chunks of this file:
 
--- a/scenes/game/ai_player.gd
+++ b/scenes/game/ai_player.gd
@@ -30,6 +30,7 @@
 ##         the top of a turn; for example, a mid-strike decision on movement.
 
 class_name AIPlayer
+extends Resource
 
 const TEST_PrepareOnly = false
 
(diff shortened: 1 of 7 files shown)


scenes/game/ai_player.gd
```diff
@@ -1,26 +1,82 @@
...
+## game.
+
+class AIPlayerState extends Resource:
+	## The underlying game object that this player state reflects.
+	var source: LocalGame.Player
 	var player_id : Enums.PlayerId
-	var deck_id
 	var life
 	var deck
+	# `kit` is the character-defining JSON, which contains the UA/XA and
```
daniel-k-taylor, line 75: let's just use deck_json then instead of defining a new term?

daniel-k-taylor: or actually, I think I use deck_def inside local_game?

scenes/game/ai_player.gd
```diff
@@ -1,26 +1,82 @@
...
+	# anything else that isn't a deck card proper
+	var kit
+	# `full_deck` is the full definitions of the actual cards in the deck,
```
daniel-k-taylor, line 78: not sure why I didn't just call this deck_list to match what it is assigned from

scenes/game/ai_player.gd
```diff
@@ -206,129 +323,77 @@ class NumberFromRangeAction:
...
+	game_state.update()
+	var possible_actions = get_boost_actions(valid_zones, limitation, ignore_costs, boost_amount)
 	return ai_policy.pick_boost_action(possible_actions, game_state)
 
-func determine_possible_turn_actions(game_logic : LocalGame, me : LocalGame.Player, opponent : LocalGame.Player):
+# TODO: Change all references to game_player below to consult game_state
+# instead? And possibly take game_state as an input if we ever need to generate
+# possible actions out of a hypothetical scenario.
+
+func determine_possible_turn_actions():
```
daniel-k-taylor, line 352: my thinking with always passing in everything and not using member variables was to make it easy to track exactly what state the AI logic is referencing and ensuring it isn't using any state outside of what should be available to the AI at that time or changing anything it shouldn't be.

That said, since I was passing in the whole game anyway it's kind of irrelevant.

I'm open to whichever design you think makes the most sense and is easiest to work with.

scenes/game/ai_player.gd
```diff
@@ -417,42 +482,43 @@ func get_change_cards_actions(_game_logic : LocalGame, me : LocalGame.Player, _o
...
 		return []
 	return [ReshuffleAction.new()]
 
-func does_boost_work(_game_logic : LocalGame, _me : LocalGame.Player, _opponent : LocalGame.Player, _card_id):
+func does_boost_work(_card_id):
 	# Examine the boost effect of the card to see if it does anything.
 	# For example, Retreat/Advance when they don't move, and there are no bonuses.
 	# Though maybe you do this just to cancel? So only:
 	# Draw when it kills you.
+	# TODO: Implement this lol
```
daniel-k-taylor, line 518: don't you know, boosts always work!

scenes/game/ai_player.gd
```diff
@@ -845,168 +903,145 @@ func determine_choose_opponent_card_to_discard_options(card_ids : Array):
...
-		var card = opponent.deck_list[i]
+		var card = game_opponent.deck_list[i]
 		if normal_only and card.definition['type'] != "normal":
 			continue
 		possible_actions.append(NameCardAction.new(card.id))
 	if can_use_own_reference:
 		# TODO: cull this down so normals aren't included twice.
-		for i in range(0, me.deck_list.size(), 2):
+		for i in range(0, game_player.deck_list.size(), 2):
 			# Skip every other card to avoid dupes.
```
daniel-k-taylor, line 984: This is wrong for some characters isn't it?
Who has odd numbered amounts of cards? At least Gordeau has a singleton.

scenes/game/ai_player.gd
```diff
@@ -845,168 +903,145 @@ func determine_choose_opponent_card_to_discard_options(card_ids : Array):
...
-	for i in range(0, opponent.deck_list.size(), 2):
+	for i in range(0, game_opponent.deck_list.size(), 2):
 		# Skip every other card to avoid dupes.
-		var card = opponent.deck_list[i]
+		var card = game_opponent.deck_list[i]
 		if normal_only and card.definition['type'] != "normal":
 			continue
 		possible_actions.append(NameCardAction.new(card.id))
 	if can_use_own_reference:
 		# TODO: cull this down so normals aren't included twice.
```
daniel-k-taylor, line 982: this todo... I think it means if both players have the same normal remove it from the list?
Don't know if this matters really though. Maybe just delete it?

test/unit/test_randomai.gd
```diff
@@ -17,31 +14,25 @@ var player1 : LocalGame.Player
...
-	ai1.game_player = player1
-	ai2 = AIPlayer.new()
-	ai2.ai_policy.free()
-	ai2.set_ai_policy(ai_policy)
-	ai2.game_player = player2
+	ai1 = AIPlayer.new(game_logic, player1, policy_type.new())
+	ai2 = AIPlayer.new(game_logic, player2, policy_type.new())
 
 func game_teardown():
+	# TODO: Move this logic into the real game so that it doesn't memory leak
```
daniel-k-taylor, line 30: As you can probably tell, I've been a bit of a mess with godot memory management.
game wrapper should probably be freed at some point in the main scene? and have teardowns to free localgame/remotegame.
Or maybe made into a node that attaches to the scene so it gets freed automatically?

scenes/game/ai_player.gd
rngesus-wept: Yeah, it's `deck_def`. Fixed.

scenes/game/ai_player.gd
rngesus-wept: I'll change this to `deck_list`, since I'm messing with this now anyway.

scenes/game/ai_player.gd
rngesus-wept: Happy Chaos has a bunch of 3-sets. I'll slap a TODO on this for now since it's not relevant to the stated purpose of this PR.

scenes/game/ai_player.gd
rngesus-wept: I think you're right; these will all eventually ingest a full game state in the form of AIGameState. I figured I'd build that out in a later PR once there's functionality and unit tests that make use of it.

scenes/game/ai_player.gd
```diff
@@ -1,26 +1,82 @@
...
-var game_state : AIGameState = AIGameState.new()
-
-var ai_policy = AIPolicyRules.new()
+var game_logic : LocalGame
+var game_state : AIGameState
+var ai_policy
+
+# TODO: Check if unused in production? Seems like this should just be
+# encapsulated in game_state.
+var game_player: LocalGame.Player
```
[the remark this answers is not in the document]
rngesus-wept: It's used now because stuff isn't being passed in anymore. But yeah, once everything is wrapped up tidily in AIGameState, these should probably go away.

scenes/game/ai_player.gd
rngesus-wept: As above, will take a closer look at this in a separate PR.

test/unit/test_randomai.gd
rngesus-wept: I'm not 100% on it myself but the unit tests do a good job of complaining when it's not being handled fully. The main thing here is that as far as I can tell, only `RefCounted` and its descendents (of which `Resource` is one) get automatically garbage-collected, and everything else has to be manually `free`d or `queue_free`d.

The TODO is mostly the idea that if you have to make multiple `free()` calls manually in the tests, that probably means that there's a `teardown()` belonging somewhere else -- maybe in `game_wrapper`? -- that should hold the top-level piece of this logic and recursively `free()` its children (like any Player objects, AI or otherwise).

daniel-k-taylor (approved):

PR #151, opened 2024-05-14
Create generic copyable resource

rngesus-wept:
Generic [Godot objects](https://docs.godotengine.org/en/stable/classes/class_object.html) don't support deep duplication, and [Godot resources](https://docs.godotengine.org/en/stable/classes/class_resource.html#class-resource) only support it if you jump through [a bunch of hoops](https://docs.godotengine.org/en/stable/classes/class_%40globalscope.html#enum-globalscope-propertyusageflags) (and maybe not even then; I kind of got that far and decided I didn't care to verify any further details). This PR provides a generic class that supports deep copy and equality checks the way you'd expect in other languages, for all those times when we need to just have a pile of structured data with no direct game engine functionality.

The name is up for bikeshedding in case we expect to ever use this outside of an AI context. (For example, it wouldn't entirely make sense to have the top-level LocalGame be an "AIResource".)

Also, move unit tests regarding basic data manipulation to their own files.


scenes/game/ai/ai_resource.gd
```diff
@@ -0,0 +1,178 @@
...
+extends Resource
+
+# AIResources usually reflect data coming from a particular source. We retain a
+# reference to that source since we'll sometimes want to sync our values to it.
+# Copies of an AIResource instead point to the original.
+var original
+
+const IGNORE_PROPERTIES = {  # Don't duplicate these properties in copy()
+	# Godot built-ins
+	'RefCounted': 1, 'script': 1, 'Built-in script': 1, 'Resource': 1,
```
daniel-k-taylor, line 31: why is this a dictionary and not an array?

scenes/game/ai/ai_resource.gd
```diff
@@ -0,0 +1,178 @@
...
+##     * Takes zero required arguments.
+##     * If all arguments have their default value, the resulting object should
+##       be as close to a blank slate as possible; for example, if there are
+##       _init behaviors that perform a sync-to-original, these should *not*
+##       happen if all values are default.
+##   * copy
+##     * This is probably just a call to copy_impl but also passes in the
+##       identity of the subclass.
+
+class_name AIResource
```
daniel-k-taylor, line 21: Maybe CopyableResource if you're planning to use this ina  lot of places?

daniel-k-taylor (changes requested):

scenes/game/ai/ai_resource.gd
rngesus-wept: Done.

scenes/game/ai/ai_resource.gd
rngesus-wept: O(1) lookup time on keys, though it doesn't really matter for such low values of n.

daniel-k-taylor (approved):

PR #157, opened 2024-07-11
Save/Load replay files on browser & native + view hand in replays

Inkie21:
"Save Replay" button appears below "Combat Log" button at game end, downloads in browser and opens a file saving dialog in native. Filename defaults to "YY-MM-DD HHhMM ryu (nick) vs ken (nick).txt", folder in native defaults to where the exe is.
"Load Replay" button in main menu prompts for a file, compatible with files made from the previous clipboard implementation.
Your hand is now visible in replays, since it makes them much more useful for game review. Saving replays should not be made possible before the game is over, since it's trivial to edit the file to flip the POV.

### DIFF
--- a/scenes/game/game.gd
+++ b/scenes/game/game.gd
@@ -5155,7 +5155,7 @@ func generate_replay_string():
 	var replay_log = {'messages': messages_list}
 	return JSON.stringify(replay_log)
 
-func replay_name():
+func get_replay_filename():
 	var filename = Time.get_datetime_string_from_system(false, true).substr(2, 14).replace(":","h")
 	filename = filename + " %s (%s) vs %s (%s).txt" % [
 		player_deck["id"],
@@ -5167,9 +5167,9 @@ func replay_name():
 func _on_save_replay_button_pressed():
 	if OS.has_feature("web"):
 		var replay_string = generate_replay_string()
-		JavaScriptBridge.download_buffer(replay_string.to_utf8_buffer(), replay_name(), "text/plain")
+		JavaScriptBridge.download_buffer(replay_string.to_utf8_buffer(), get_replay_filename(), "text/plain")
 	else:
-		file_dialog.current_file = replay_name()
+		file_dialog.current_file = get_replay_filename()
 		file_dialog.visible = true
 
 func _on_file_dialog_file_selected(path):
--- a/scenes/game/game.gd
+++ b/scenes/game/game.gd
@@ -41,6 +41,9 @@ const LocationInfoButtonPair = preload("res://scenes/game/location_infobutton_pa
 @onready var emote_dialog : EmoteDialog = $EmoteDialog
 @onready var modal_dialog : ModalDialog = $ModalDialog
 
+@onready var save_replay_button = $PlayerZones/SaveReplayButton
+@onready var file_dialog = $FileDialog
+
 const OffScreen = Vector2(-1000, -1000)
 const ChoiceCopyIdRangeStart = 70000
 const RevealCopyIdRangestart = 80000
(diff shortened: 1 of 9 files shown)


scenes/game/game.gd
```diff
@@ -5141,11 +5150,32 @@ func _on_combat_log_button_pressed():
...
 
-func _on_combat_log_replay_button_pressed():
+func generate_replay_string():
 	var messages_list = [starting_message.duplicate()] + game_wrapper.get_message_history()
 	var replay_log = {'messages': messages_list}
-	var replay_log_string = JSON.stringify(replay_log)
-	DisplayServer.clipboard_set(replay_log_string)
+	return JSON.stringify(replay_log)
+
+func replay_name():
```
daniel-k-taylor, line 5158: let's call this get_replay_filename()

scenes/menu/main_menu.gd
```diff
@@ -430,11 +442,18 @@ func _on_modal_list_observe_match_pressed(row_index):
...
 
 func _on_settings_button_pressed():
 	settings_window.visible = true
 
+func load_replay(data):
+	var replay_data = JSON.parse_string(data[0])
+	_on_observe_game_started(replay_data, true)
+
+func _on_file_dialog_file_selected(path):
+	load_replay([FileAccess.get_file_as_string(path)])
```
daniel-k-taylor, line 459: Did you test canceling the dialog to make sure that works?

scenes/menu/main_menu.gd
```diff
@@ -430,11 +442,18 @@ func _on_modal_list_observe_match_pressed(row_index):
...
+	if OS.has_feature("web"):
+		window.input.click()
+	else:
+		file_dialog.visible = true
 
 func _on_settings_button_pressed():
 	settings_window.visible = true
 
+func load_replay(data):
+	var replay_data = JSON.parse_string(data[0])
```
daniel-k-taylor, line 455: let's handle if this JSON doesn't parse correctly and you get back null.
Maybe throw up an error message?

scenes/game/game.gd
```diff
@@ -625,7 +633,7 @@ func spawn_deck(deck_id,
 		var logic_card : GameCard = card_db.get_card(card.id)
 		var image_path = get_card_image_path(deck_id, logic_card)
 		var new_card = create_card(card.id, logic_card.definition, image_path, card_back_image, deck_card_zone, is_opponent)
-		if observer_mode:
+		if observer_mode and not replay_mode:
```
daniel-k-taylor, line 636: Is this enough to see the opponent cards too? I don't think it is?

daniel-k-taylor (changes requested):

scenes/game/game.gd
Inkie21: Done

scenes/menu/main_menu.gd
Inkie21: Yeah, this only gets called when a file is confirmed, and clicking the Load Replay button only pops out the dialog, so nothing happens if you cancel.

scenes/menu/main_menu.gd
Inkie21: Done

scenes/game/game.gd
Inkie21: It isn't, I don't think this would be good to do. The primary purpose of replays is to review decisions that were made in a given game state, with the limited information that was available at the time. Showing the opponent's hand would inevitably pollute that assessment, even if we try not to take it into account.

daniel-k-taylor: Maybe we should make this an option?
I could see wanting to see both hands if you're casting the game or something.
Or if you want to know why your opponent played what they did at the time.

daniel-k-taylor (approved):

PR #160, opened 2024-08-03
three new emotes. more emote transparency work

SeaGriff:
last one of these i think


assets/icons/emotes/elpheltragingdemon.webp.import
```diff
@@ -0,0 +1,34 @@
+[remap]
```
daniel-k-taylor, line 1: can you delete your old ones? like the jpg version

daniel-k-taylor (changes requested):

PR #165, opened 2024-09-24
Update to Godot 4.3

Mach131:


scenes/game/game.gd
```diff
@@ -10,7 +10,6 @@ const Test_StartWithGauge = false
 
 const CardBaseScene = preload("res://scenes/card/card_base.tscn")
 const CardBase = preload("res://scenes/card/card_base.gd")
-const Enums = preload("res://scenes/game/enums.gd")
 const CardPopout = preload("res://scenes/game/card_popout.gd")
```
daniel-k-taylor, line 13: can we actually get rid of all the ones with .gd if we make sure we add class_name to every file?

daniel-k-taylor (approved):

scenes/game/game.gd
Mach131: There doesn't seem to be much of a way around doing so for "tscn" files, but will do for all the "gd"s

project.godot
```diff
@@ -12,7 +12,7 @@ config_version=5
 
 config/name="FightingCards"
 run/main_scene="res://scenes/main.tscn"
-config/features=PackedStringArray("4.1", "GL Compatibility")
```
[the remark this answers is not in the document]
Mach131: yup, seems good at a glance

daniel-k-taylor (approved):

PR #170, opened 2025-02-18
Celinka

daniel-k-taylor:


data/card_definitions.json
```diff
@@ -3419,6 +3419,339 @@
...
+		"power": 3,
+		"armor": 0,
+		"guard": 0,
+		"effects": [
+			{
+				"timing": "after",
+				"effect_type": "armorup_per_continuous_boost",
+				"amount": 2,
+				"and": {
+					"effect_type": "seal_continuous_boosts"
```
Mach131, line 3532: idle thought, do cards given to you by the opponent get sealed (i.e. faust's afro)? i think they should be able to but not sur if we've had to worry about that already

data/decks/celinka.json
```diff
@@ -0,0 +1,205 @@
...
+	"has_sealed_area": true,
+	"exceed_cost": 5,
+	"exceed_cost_reduced_by": [{ "reduction_type": "transform_discount" }],
+	"ability_effects": [
+		{
+			"timing": "on_prepare",
+			"effect_type": "self_discard_choose",
+			"optional": true,
+			"amount": 3,
+			"destination": "boost_as_hidden_powerup_and_seal"
```
Mach131, line 14: probably not relevant for the purposes of implementing celinka, but it may be worth having a more generic way to handle face-down boosts eventually

scenes/game/local_game.gd
```diff
@@ -7832,6 +7947,23 @@ func handle_strike_effect(card_id : int, effect, performing_player : Player):
...
+					"display_name": "The Gaki Boost",
+					"effects": [
+						{
+							"timing": "during_strike",
+							"effect_type": "powerup",
+							"amount": 1
+						}
+					]
+				}
+				performing_player.play_replacement_boosts(card_ids, replacement_boost)
```
Mach131, line 7966: iirc there is a cursed interaction with facedown boosts where they lose their effects if the text defining them leaves play (that is, if celinka exceeds while some facedown boosts are still around); i'm not 100% confident this is true though and i would be willing to pretend it's not because it sounds extremely painful to deal with

Mach131 (dismissed):

scenes/game/local_game.gd
daniel-k-taylor: that sounds dumb, so I'm going to say they stay.

PR #172, opened 2025-02-23
Emogine

daniel-k-taylor:


data/card_definitions.json
```diff
@@ -7059,6 +7059,347 @@
...
+				"choice": [
+					{
+						"effect_type": "boost_then_sustain",
+						"valid_zones": [ "hand" ],
+						"limitation": "transform",
+						"sustain": false
+					},
+					{
+						"effect_type": "close",
+						"amount": 2
```
Mach131, line 7327: Does this force you to choose one of the effects? It doesn't seem like this, but if so we may need to force them to close if they don't have a valid card in hand to transform

data/decks/emogine.json
```diff
@@ -0,0 +1,215 @@
...
+			"timing": "set_strike",
+			"condition": "was_wild_swing",
+			"effect_type": "reveal_strike",
+			"and": {
+				"effect_type": "choice",
+				"choice": [
+					{ "effect_type": "spend_life", "amount": 1, "and": {
+						"effect_type": "replace_wild_swing",
+						"previous_attack_to": "gauge"
+					} },
```
Mach131, line 18: This might not be technically correct if Emogine's deck is empty when setting a wild swing (since the "top of deck" isn't there), though maybe there's an errata for this i'm not aware of

scenes/menu/char_select.tscn
```diff
@@ -787,6 +788,11 @@ layout_mode = 2
 char_id = "celinka"
 portrait_texture = ExtResource("89_d1xn5")
 
+[node name="Emogine" parent="CenterContainer/SCCharacterSelect/Rows/Row1" instance=ExtResource("6_3lmwy")]
```
Mach131, line 791: if you haven't already it's probably getting to the point where we might want a second row for S2

data/card_definitions.json
daniel-k-taylor: Sounds like no, i added pass

data/decks/emogine.json
daniel-k-taylor: I'll add the min cards in deck condition to it with 1

scenes/menu/char_select.tscn
daniel-k-taylor: Maybe next character

PR #175, opened 2025-03-03
Remiliss

daniel-k-taylor:


data/card_definitions.json
```diff
@@ -22796,6 +22796,342 @@
...
+					"effect_type": "attack_copy_gauge_or_transform_becomes_ex",
+					"hide_effect": true
+				},
+				{
+					"NOTE": "This timing is when the stat boosts are updated so it is actually EX",
+					"timing": "during_strike",
+					"condition": "copy_of_attack_in_zones",
+					"condition_zones": ["gauge", "transform"],
+					"effect_type": "attack_is_ex",
+					"override_description": "If copy of attack is in Gauge/Transform, attack is EX"
```
Mach131, line 23036: quick check - if you reveal Consuption and pay with another Consumption from gauge, does it remain EX? (it should, according to some people I asked in VC)

data/card_definitions.json
```diff
@@ -22796,6 +22796,342 @@
...
+			"boost_type": "immediate",
+			"force_cost": 0,
+			"cancel_cost": -1,
+			"display_name": "Engineered Destruction",
+			"effects": [
+				{
+					"timing": "immediate",
+					"effect_type": "draw",
+					"amount": 4,
+					"reveal": true
```
Mach131, line 23079: this seems to be worded as top of deck - does this work if you have less than 4 cards left?

daniel-k-taylor: I'll have to use strike_x probably to set it to 4 or # left in deck, then use that for the other effects

data/card_definitions.json
daniel-k-taylor: okay, added

PR #177, opened 2025-03-07
Custom bonus effects

Mach131:


scenes/game/local_game.gd
```diff
@@ -4852,6 +4860,10 @@ func handle_strike_effect(card_id : int, effect, performing_player : Player):
 			events += performing_player.remove_from_continuous_boosts(card, "overdrive")
 		"add_hand_to_gauge":
 			events += performing_player.add_hand_to_gauge()
+		"add_opponent_strike_to_gauge":
+			# please only do this if the opponent is stunned
```
daniel-k-taylor, line 4864: Why only if stunned? Because if card1 is activating and you remove card 2 then... who knows???
Maybe we should just throw here or something.
Though to help people debug customs, maybe send an error event so a dialog pops up explaining what is wrong.

daniel-k-taylor (changes requested):

daniel-k-taylor (changes requested):
Please add tests for the new effects

scenes/game/local_game.gd
Mach131: I actually wrote this before I fully figured out how the effect would work; I was worried about whether cleanup effects would still go off but the remove from strike function handles that fine. I'll keep the error event in mind for later, maybe when i go through and try to make documentation

daniel-k-taylor (approved):

PR #194, opened 2025-06-08
Added various custom elements.

Sompursone:
The commits cover it further, but to summarize, added player variables set_starting_face_attack + starting_face_attack_id, condition opponent_at_min_range, and effects name_speed and opponent_discard_speed_or_reveal


data/decks/test_custom.json
```diff
@@ -0,0 +1,191 @@
+{
```
daniel-k-taylor, line 1: it should be really easy to add a custom character, in exceed_test.gd, what you're looking for is default_game_setup()
You can see that this is how the deck is loaded:
var default_deck = CardDataManager.get_deck_from_str_id(who_am_i())

There are a few ways to do this, the easiest is probably to just do like:
var character_identity = who_am_i()
if typeof(character_identity) == TYPE_DICTIONARY:
   default_deck = character_identity
else:
   ... what we did before with the carddatamanager.

daniel-k-taylor: Then in your test case, just return the full character deck json in the who_am_i func
I'd prefer this over adding a bunch of test custom jsons into the decks list
and if you want to make it shorter, just only give them like a bunch of 1 type of card, so you won't have to include more images and stuff in the json

daniel-k-taylor (changes requested):

daniel-k-taylor (approved):

PR #195, opened 2025-06-13
[Condition] Was Pushed or Pulled

vgma2:
Added conditions to check if you or the opponent had been pushed/pulled by the other player

### DIFF
--- a/scenes/core/player.gd
+++ b/scenes/core/player.gd
@@ -299,6 +299,8 @@ var starting_hand_size_bonus : int
 var draw_at_end_of_turn : int
 var pre_strike_movement : int
 var moved_self_this_strike : bool
+var was_pushed_or_pulled_this_strike : bool
+var spaces_forced_moved_this_strike : int
 var moved_past_this_strike : bool
 var spaces_moved_this_strike : int
 var spaces_moved_or_forced_this_strike : int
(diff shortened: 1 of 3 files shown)


documents/Conditions.md
```diff
@@ -0,0 +1,24 @@
...
+The conditions listed in this document will contain conditions that are used by base game characters, as well as ones that are only used by custom characters.
+This list is currently incomplete and will be gradually updated.
+
+`was_pushed_or_pulled_this_strike`
+Returns true if your character had been pushed or pulled during this strike.
+(Does not count if the you did not move from being pushed/pulled)
+
+`opponent_was_pushed_or_pulled_this_strike`
+Returns true if the opponent had been pushed or pulled during this strike.
+(Does not count if the opponent did not move from being pushed/pulled)
```
vgma2, line 14: `was_pushed_or_pulled_this_strike` and `opponent_was_pushed_or_pulled_this_strike` seems like they would be redundant with the two conditions below.  Do we have a use case for these where we might want to use them, compared to the ones below?  Otherwise, I will probably remove these two conditions.

daniel-k-taylor: hmm, probably not?  Seems like you should just keep the one where you pass in an amount?

scenes/core/player.gd
```diff
@@ -299,6 +299,8 @@ var starting_hand_size_bonus : int
 var draw_at_end_of_turn : int
 var pre_strike_movement : int
 var moved_self_this_strike : bool
+var was_pushed_or_pulled_this_strike : bool
+var spaces_forced_moved_this_strike : int
```
daniel-k-taylor, line 303: you never clear this anywhere?

daniel-k-taylor (changes requested):
I wonder if we also want an example effect checked in somewhere?

scenes/core/player.gd
vgma2: It's done in lines 725-726 of local_game.gd.  Are there other places where values like these need to be cleared?

daniel-k-taylor: oh, no, that's good. I thought that was the init function instead

daniel-k-taylor (approved):

